PR #10, opened 2018-03-20
Expose Job jid

wezm:
For tracing in my application I want to be able to return the jid of the job enqueued as the result of a request. This small change makes `jid` available on `Job`.


jonhoo:
Ah, yes, good catch!

jonhoo:
(Being) published in 0.9.3!
I renamed the method to `id`.

PR #15, opened 2018-03-22
Rework consumer type to use function-like-object via a trait

icefoxen:
…instead of closures.

It seems to actually work quite smoothly, and you can still pass closures into it to create job runners.  So, this is strictly more powerful, I think.

One potential inconvenience is the trait takes `&self` instead of `&mut self`; changing that would require making callbacks an `Arc<FnvHashmap<..., Mutex<JobRunner>>>` instead of just `Arc<FnvHashmap<...>>`.  I dunno which way is the best.  You can work around it when you implement `JobRunner` just by making all its fields that need to mutate interior-mutable with `Mutex` or such, so it's not a killer.  Might even be nicer this way so you don't *have* to lock stuff when you don't need to.

I sorta feel like having a thread pool internal to the `Consumer` is a bit wonky, and it should be possible to have a `Consumer` that defines operations and a `ConsumerRunner` that may have different strategies for running it; that way we could maybe loosen some `Send` and `Sync` constraints on `JobRunner`'s.  I'm trying to use this where each job needs to use tokio to do some async communication watsit, so each `JobRunner` type includes a thread pool anyway, and it feels a bit redundant.  This isn't a design issue I know how to fix yet though.

### DIFF
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -13,6 +13,28 @@ const STATUS_RUNNING: usize = 0;
 const STATUS_QUIET: usize = 1;
 const STATUS_TERMINATING: usize = 2;
 
+/// A trait for anything that can handle a Faktory job.
+///
+/// Unfortunately a functor-type object like this is far more convenient
+/// and flexible than a closure.
+///
+/// TODO: It would be nice to loosen the Send+Sync and immutable constraints some...
+/// Probably can't as long as we assume this has to be runnable in a thread pool, though.
+pub trait JobRunner: Send + Sync {
+    type Error;
+    fn run(&self, job: Job) -> Result<(), Self::Error>;
+}
+
+impl<E, T> JobRunner for T
+    where T: Fn(Job) -> Result<(), E> + Send + Sync {
+    type Error = E;
+    fn run(&self, job: Job) -> Result<(), Self::Error> {
+        self(job)
+    }
+}
+
+type BoxedJobRunner<E> = Box<JobRunner<Error=E>>;
+
 /// `Consumer` is used to run a worker that processes jobs provided by Faktory.
 ///
 /// # Building the worker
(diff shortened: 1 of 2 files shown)


icefoxen:
I tried but just using boxed closures but the type inference just got too hairy for me to handle.  I'm curious now so I'm going to give it another go.

The downside to that is that you can't implement `Fn()` on your own types, so you *must* use closures.  I personally find this gets less convenient the bigger a project gets, but YMMV.

jonhoo:
Ah, I see what you mean. If you want to provide things that aren't closures, I guess the way to do that if we stick with the `Fn` bound would be to have a `move` closure that calls into your other type. I think of job handlers as similar to Rocket's [route handlers](https://api.rocket.rs/rocket/struct.Route.html#method.new), which are also basically [just `Fn`](https://api.rocket.rs/rocket/type.Handler.html). I guess I have a preconceived notion that *most of the time*, people will just want to "do some stuff" when a job of a particular type comes in, and closures are a convenient way of expressing that. We could have *two* methods I suppose, one that takes a `Fn` and one that takes a trait type, and then implement that `JobRunner` trait for `Fn`...

jonhoo:
I don't want to push more work in your lap, but I do think it's worthwhile to do this "properly" if we're first going to make a change :) And taking inspiration from the (at least in my mind) very similar Rocket model is probably a good place to start.  I'd be inclined to do a `Box<Fn>` approach over a trait scheme unless we match Rocket just because it is (/should be) a relatively straightforward transformation from what we currently have, while still serving your use-case. That is, I do think you can still share connections and stuff with a closure (you'd just move the state into the closure, and then you can access it inside the closure in essentially the same way as `&self` would), but I also agree it may be a less than ideal interface for more advanced state management.

icefoxen:
No worries, I gotta get back to other projects anyway so I'll be off of this for a little while.  I tried to make this minimal-impact, from the user's side nothing changes.

I just make a `Box<Fn>` version and it came out nicer than I expected, let me clean it up a bit and push another PR so we can compare the two.

PR #16, opened 2018-03-22
Pure Box<Fn> type version of PR #15

icefoxen:


src/consumer/mod.rs
```diff
@@ -178,21 +180,18 @@ impl<F> ConsumerBuilder<F> {
...
-impl<F, E> ConsumerBuilder<F>
-where
-    F: Fn(Job) -> Result<(), E> + Send + Sync + 'static,
-{
     /// Register a handler function for the given job type (`kind`).
     ///
     /// Whenever a job whose type matches `kind` is fetched from the Faktory, the given handler
     /// function is called with that job as its argument.
-    pub fn register<K>(&mut self, kind: K, handler: F) -> &mut Self
+    pub fn register<K, J>(&mut self, kind: K, handler: J) -> &mut Self
```
jonhoo: Can we make this `H` for `Handler`?

src/consumer/mod.rs
```diff
@@ -345,11 +343,10 @@ where
     }
 }
 
-impl<S, E, F> Consumer<S, F>
+impl<S, E> Consumer<S, E>
 where
     S: Read + Write + Reconnect + Send + 'static,
-    E: StdError,
-    F: Fn(Job) -> Result<(), E> + Send + Sync + 'static,
+    E: StdError + 'static,
```
jonhoo: Huh, curious that `E` needs to be `'static` here now. I don't think it's an issue, just curious as to why.

jonhoo:
This looks good! See the minor inline comments.

src/consumer/mod.rs
icefoxen: Sure; it was `J` for `JobHandler`.  :man_facepalming:

src/consumer/mod.rs
icefoxen: Good question!  I don't remember, I'll check.

icefoxen: If it's not `'static` I get the following:

```
   --> src/consumer/mod.rs:422:20
    |
346 | impl<S, E> Consumer<S, E>
    |         - help: consider adding an explicit lifetime bound `E: 'static`...
...
422 |                 Ok(thread::spawn(move || {
    |                    ^^^^^^^^^^^^^
    |
note: ...so that the type `[closure@src/consumer/mod.rs:422:34: 431:18 status:std::sync::Arc<std::sync::atomic::AtomicUsize>, w:consumer::Consumer<S, E>, worker:usize, queues:std::vec::Vec<std::string::String>]` will meet its required lifetime bounds
   --> src/consumer/mod.rs:422:20
    |
422 |                 Ok(thread::spawn(move || {
    |                    ^^^^^^^^^^^^^

```

However I'm not quite sure how/whether to restructure it to fix that...

jonhoo: Ah, interesting -- so it's because the error may need to be propagated back across the thread boundary. `'static` seems fine then :)

jonhoo:
Published in 0.10! Thanks again :D

PR #23, opened 2021-04-25
Add commands to pause and resume queues

Syfaro:
Hi there! [Faktory 1.5](https://github.com/contribsys/faktory/blob/master/Changes.md#150) added commands to pause and resume queues, this PR adds these commands to this library. I believe I have implemented them correctly and added useful tests, but please feel free to suggest improvements.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/23)

### DIFF
--- a/src/producer/mod.rs
+++ b/src/producer/mod.rs
@@ -1,4 +1,4 @@
-use crate::proto::{self, Client, Info, Job, Push};
+use crate::proto::{self, Client, Info, Job, Push, QueuePause, QueueResume};
 use failure::Error;
 use serde_json;
 use std::io::prelude::*;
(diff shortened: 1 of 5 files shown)


jonhoo:
Ah, this is great, thank you! The one thing I'd like to see improved is for the "real" test to actually check that the queues are indeed paused and resumed. Could you include actually enqueueing some jobs on the relevant queues, and then check that they're only made available to workers when the queue isn't paused?

Syfaro:
Sure! I've updated the real test to pause and resume a queue and ensure a job is only made available at the right time.

src/producer/mod.rs
```diff
@@ -116,6 +116,20 @@ impl<S: Read + Write> Producer<S> {
             .read_json()
             .map(|v| v.expect("info command cannot give empty response"))
     }
+
+    /// Pause queues.
```
jonhoo, line 120:
```suggestion
    /// Pause the given queues.
```

src/producer/mod.rs
```diff
@@ -116,6 +116,20 @@ impl<S: Read + Write> Producer<S> {
...
+    pub fn queue_pause<T: AsRef<str>>(&mut self, queues: &[T]) -> Result<(), Error> {
+        self.c
+            .issue(&QueueControl::new(QueueAction::Pause, queues))?
+            .await_ok()
+    }
+
+    /// Resume queues.
```
jonhoo, line 127:
```suggestion
    /// Resume the given queues.
```

jonhoo (approved):
Looks great, thank you!

jonhoo:
Publishing as 0.11.1 :+1:

PR #24, opened 2021-07-02
perform_in and perform_at

worst:
Added functionality for scheduled jobs (via the `at` field in `Job`), which to the best of what I can tell, was not currently implemented.

I named these methods based on the Ruby/sidekiq client. `perform_in` and `perform_at` take a duration for how long in the future should be scheduled for and a specific time to schedule a job to be enqueued, respectively.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/24)

### DIFF
--- a/src/producer/mod.rs
+++ b/src/producer/mod.rs
@@ -120,7 +120,6 @@ impl<S: Read + Write> Producer<S> {
         self.enqueue(job)
     }
 
-
     /// Enqueue the given job on the Faktory server to be performed at a certain time
     ///
     /// Returns `Ok` if the job was successfully queued by the Faktory server.
@@ -155,8 +154,6 @@ impl<S: Read + Write> Producer<S> {
     pub fn enqueue_at(&mut self, job: Job, at: DateTime<Utc>) -> Result<(), Error> {
         self.perform_at(job, at)
     }
-
-    
 
     /// Retrieve information about the running server.
     ///
--- a/src/producer/mod.rs
+++ b/src/producer/mod.rs
@@ -1,8 +1,11 @@
 use crate::proto::{self, Client, Info, Job, Push, QueueAction, QueueControl};
+use chrono::Duration;
 use failure::Error;
 use serde_json;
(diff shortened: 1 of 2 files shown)


jonhoo:
I'm not sure I follow why all these extra methods are needed when the `at` field on `Job` is already public? Do they add all that much to the interface?

worst:
Perhaps I'm trying to force meme too much of the Ruby API into this (I'm coming from using sidekiq for years, and that API is reproduced in the "official" Faktory Ruby library), but, typically, my workflow is to build a Job (arguments, etc.), and but scheduling it is a separate operation.

Maybe it doesn't make sense for the Rust library though?

jonhoo:
I think I'd actually rather see convenience methods on `Job` itself to set its timing in a nicer way, but even that doesn't seem like it'll make the ergonomic much nicer? Let's assume we have some `job: Job` we want to schedule; compare:

```rust
job.at = Some(Utc.ymd(2022, 11, 14).and_hms(8, 9, 10));
faktory.enqueue(job);
```

```rust
job.perform_at(Utc.ymd(2022, 11, 14).and_hms(8, 9, 10));
faktory.enqueue(job);
```

```rust
faktory.perform_at(job, Utc.ymd(2022, 11, 14).and_hms(8, 9, 10));
```

It's true that the last one is one line shorter, but it also hides the fact that all this is really doing is configuring the `job` (and indeed _overwriting_ `job.at`), so I'm not sure if it's an overall win? Maybe we could improve the documentation on `Job` itself to make it clearer how to set a future scheduling date for a job?

worst:
Sorry about delayed reply.

I spent some time thinking about this and I'm wondering if it's a better idea to create a new crate that is focused on "ergonomics" (or whatever) whereas the current crate would be a lower level "driver" type of thing?

I have some ideas on macros and some traits that, while I think can provide some ergonomics, probably do not belong in this crate.

Anyways, I'm going to close out this pull request, because I think you are probably right on this one.

PR #25, opened 2021-09-17
Fix clippy warnings

ttdonovan:
Fixed clippy warnings and replaced code with recommended suggestion.

* `cargo clippy`
* `cargo fmt`

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/25)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,7 @@
 jobs:
  - template: default.yml@templates
    parameters:
-     minrust: 1.42.0
+     minrust: 1.51
  - job: integration
    displayName: cargo test (real)
    pool:
--- a/src/consumer/mod.rs
+++ b/src/consumer/mod.rs
@@ -535,7 +535,7 @@ where
                     .into_iter()
                     .map(|w| w.join().unwrap())
                     .collect::<Result<Vec<_>, _>>()
-                    .and_then(|_| Err(e))
+                    .and(Err(e))
             }
         }
     }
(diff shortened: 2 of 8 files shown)


ttdonovan:
@jonhoo could you provide some insight into why the minimum support Rust version is 1.42?

jonhoo:
Thanks for doing this!

@ttdonovan Purely because I went for the lowest supported version at the time of implementation. I'd be happy to bump it, though I'd prefer to stay with a version that's at least six months old (~4 releases, so <=1.51).

ttdonovan:
That seems reasonable `strip_prefix` has been in stable since 1.45 so hopefully a bump in the Rust version will get CI passing again.

https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix

Would it also be possible to add the a step in the Azure pipeline to check `cargo fmt` and `cargo clippy`?

jonhoo:
> That seems reasonable `strip_prefix` has been in stable since 1.45 so hopefully a bump in the Rust version will get CI passing again.

You should be able to do that by just modifying this line:

https://github.com/jonhoo/faktory-rs/blob/5d5308c7958ec6d53e4ab6af621a1cc9115ea9af/azure-pipelines.yml#L4

> Would it also be possible to add the a step in the Azure pipeline to check `cargo fmt` and `cargo clippy`?

There's already a step for that — "Style linting" :)

jonhoo (approved):
Looks good to me, thanks!

Issue #27, opened 2022-04-12
Open to replacing `failure` with `thiserror`?


kitified:
Hey there! I'm starting a new project using this library.

Now that `failure` [is deprecated](https://github.com/rust-lang-deprecated/failure#failure---a-new-error-management-story), and the ecosystem is moving towards a system where libraries report implementations of `Error`, would you be open to a PR replacing usage of `failure` with proper `Error` implementations?

jonhoo:
Absolutely!

kitified:
@jonhoo awesome, thanks! I put in a PR here: https://github.com/jonhoo/faktory-rs/pull/28

kitified:
@jonhoo also, I was looking over the other issues and saw this:

> I'm also looking for maintainers if anyone using Faktory on a regular basis is looking to step up to the plate

My company is building out at least one large-ish rust system that uses faktory for job dispatch (and we already use faktory heavily in other languages), so I will likely be trying to upstream multiple PRs as part of that over the coming weeks/months. 

I'm open to stepping up to the plate here- happy to talk more about what that'd look like for you and how I can help!

jonhoo:
@kitified That'd be amazing! I've added you as a maintainer of the GitHub project. I'll hold off on the crates.io release privileges for the time being, but as you ramp up on the project I'm happy to share that too :) Even though you have permissions now, please do keep submitting PRs and I'll do my best to give reviews in a timely fashion!

kitified:
Sounds great, will do!

Issue #8, opened 2018-03-18
Use error type instead of using catch all failure::Error in Producer::con


wezm:
Per the [failure guidance](https://boats.gitlab.io/failure/use-error.html#caveats-on-this-pattern) the `failure::Error` type is typically only recommended when prototyping or for "applications" that won't handle the error. It would be better to define a custom error type and 'derive Fail' for it. 

I see there are some parts of the code using a custom error type but others are falling back on Error.

If this sounds reasonable I'm happy to make this change.

jonhoo:
For the various methods on `Producer`, I completely agree that a custom error type would be useful. For `Consumer::run*` it's less clear that the user of the library can do anything useful on error, so I'm less concerned. Happy to take a look at a PR!

jonhoo:
@wezm still want to take a stab at this too? I'd be happy to review and merge!

jonhoo:
@wezm still want to take a stab at this too? I'd be happy to review and merge!

PR #28, opened 2022-04-19
Transition to `std::error::Error`; minor cleanup

kitified:
_Closes https://github.com/jonhoo/faktory-rs/issues/8
Closes https://github.com/jonhoo/faktory-rs/issues/27_

**This is a breaking change.**

Removes `failure`, and in its place implements errors as standard enums compatible with `std::error::Error`, powered by `thiserror`.
Also includes some minor cleanup and reorganization:
- The top-level crate now exposes `faktory::error::ProtocolError`, which is the old `FaktoryError` type, and `faktory::error::Error`, which is all possible errors returned by the crate (one of which is `faktory::error::ProtocolError`).
- Errors under `producer` and `consumer` now return the top-level `faktory::error::Error` type (this solves #8).
- I also made `Job.failure` into `pub`, instead of `pub(crate)`; I did this because clippy was complaining that it's unused. If you disagree happy to change it back.
- I made a couple other minor edits based on clippy recommendations.

---
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/28)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -45,7 +45,7 @@ pub enum Error {
     /// Indicates an error in the underlying TLS stream.
     #[cfg(feature = "tls")]
     #[error("underlying tls stream")]
-    TlsStream(#[from] native_tls::Error),
+    TlsStream(#[source] native_tls::Error),
 }
 
 /// Errors specific to connection logic.
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,49 +1,90 @@
-use thiserror::Error;
+//! Enumerates all errors that this crate may return.
+//!
+//! [Error] is the top level error enum.
+//! Most consumers should only need to interact with this type.
+//! This is also where more generic errors such as IO errors are placed,
+//! whereas the more specific errors ([Connection] and [Protocol]) are
+//! related to logic.
+//!
+//! [Connect] describes errors specific to the connection logic, for example
+//! version mismatches or an invalid URL.
+//!
+//! [Protocol] describes lower-level errors relating to communication
+//! with the faktory server. Typically, [Protocol] errors are the result
+//! of the server sending a response this client did not expect.
 
-use crate::proto::ConnectError;
+use thiserror::Error;
 
 /// The set of observable errors when interacting with a Faktory server.
 #[derive(Debug, Error)]
-#[allow(clippy::manual_non_exhaustive)]
+#[non_exhaustive]
 pub enum Error {
     /// The connection to the server, or one of its prerequisites, failed.
-    #[error("connection error: {0}")]
-    Connect(#[from] ConnectError),
+    #[error("connection")]
+    Connect(#[from] Connect),
 
-    /// Underlying io layer errors.
+    /// Underlying IO layer errors.
+    ///
     /// These are overwhelmingly network communication errors on the socket connection to the server.
-    #[error("underlying i/o: {0}")]
-    GenericIO(#[from] std::io::Error),
+    #[error("underlying IO")]
+    IO(#[from] std::io::Error),
 
     /// Application-level errors.
+    ///
     /// These generally indicate a mismatch between what the client expects and what the server expects.
-    #[error("protocol: {0}")]
-    Protocol(#[from] ProtocolError),
+    #[error("protocol")]
+    Protocol(#[from] Protocol),
 
     /// Faktory payloads are JSON encoded.
+    ///
     /// This error is one that was encountered when attempting to deserialize a response from the server.
     /// These generally indicate a mismatch between what the client expects and what the server provided.
-    #[error("deserialize payload: {0}")]
+    #[error("deserialize payload")]
     DeserializePayload(#[from] serde_json::Error),
 
     /// Indicates an error in the underlying TLS stream.
     #[cfg(feature = "tls")]
-    #[error("underlying tls stream: {0}")]
+    #[error("underlying tls stream")]
     TlsStream(#[from] native_tls::Error),
+}
 
-    // We're going to add more error types in the future
-    // https://github.com/rust-lang/rust/issues/44109
-    //
-    // This forces users to write pattern matches with a catch-all `_` arm.
-    #[error("unreachable")]
-    #[doc(hidden)]
-    __Nonexhaustive,
+/// Errors specific to connection logic.
+#[derive(Debug, Error)]
+pub enum Connect {
+    /// The scheme portion of the connection address provided is invalid.
+    #[error("unknown scheme: {scheme}")]
+    BadScheme {
+        /// The scheme that was provided in the connection address.
+        scheme: String,
+    },
+
+    /// The provided connection address does not contain a hostname.
+    #[error("no hostname given")]
+    MissingHostname,
+
+    /// The server requires authentication, but none was provided.
+    #[error("server requires authentication")]
+    AuthenticationNeeded,
+
+    /// The server expects a different protocol version than this library supports.
+    #[error("server version mismatch (theirs: {theirs}, ours: {ours})")]
+    VersionMismatch {
+        /// The protocol version this library supports.
+        ours: usize,
+
+        /// The protocol version the server expects.
+        theirs: usize,
+    },
+
+    /// The connection address provided was not able to be parsed.
+    #[error("parse URL")]
+    ParseUrl(#[source] url::ParseError),
 }
 
 /// The set of observable application-level errors when interacting with a Faktory server.
 #[derive(Debug, Error)]
-#[allow(clippy::manual_non_exhaustive)]
-pub enum ProtocolError {
+#[non_exhaustive]
+pub enum Protocol {
     /// The server reports that an issued request was malformed.
     #[error("request was malformed: {desc}")]
     Malformed {
@@ -80,35 +121,27 @@ pub enum ProtocolError {
         /// The relevant bytes sent by the server.
         bytes: Vec<u8>,
     },
-
-    // We're going to add more error types in the future
-    // https://github.com/rust-lang/rust/issues/44109
-    //
-    // This forces users to write pattern matches with a catch-all `_` arm.
-    #[error("unreachable")]
-    #[doc(hidden)]
-    __Nonexhaustive,
 }
 
-impl ProtocolError {
+impl Protocol {
     pub(crate) fn new(line: String) -> Self {
         let mut parts = line.splitn(2, ' ');
         let code = parts.next();
         let error = parts.next();
         if error.is_none() {
-            return ProtocolError::Internal {
+            return Protocol::Internal {
                 msg: code.unwrap().to_string(),
             };
         }
         let error = error.unwrap().to_string();
 
         match code {
-            Some("ERR") => ProtocolError::Internal { msg: error },
-            Some("MALFORMED") => ProtocolError::Malformed { desc: error },
-            Some(c) => ProtocolError::Internal {
+            Some("ERR") => Protocol::Internal { msg: error },
+            Some("MALFORMED") => Protocol::Malformed { desc: error },
+            Some(c) => Protocol::Internal {
                 msg: format!("{} {}", c, error),
             },
-            None => ProtocolError::Internal {
+            None => Protocol::Internal {
                 msg: "empty error response".to_string(),
             },
         }
--- a/src/error.rs
+++ b/src/error.rs
@@ -50,6 +50,7 @@ pub enum Error {
 
 /// Errors specific to connection logic.
 #[derive(Debug, Error)]
+#[non_exhaustive]
 pub enum Connect {
     /// The scheme portion of the connection address provided is invalid.
     #[error("unknown scheme: {scheme}")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -37,10 +37,10 @@ pub enum Error {
 
     /// Faktory payloads are JSON encoded.
     ///
-    /// This error is one that was encountered when attempting to deserialize a response from the server.
+    /// This error is one that was encountered when attempting to serialize or deserialize communication with the server.
     /// These generally indicate a mismatch between what the client expects and what the server provided.
-    #[error("deserialize payload")]
-    DeserializePayload(#[source] serde_json::Error),
+    #[error("serialization")]
+    Serialization(#[source] serde_json::Error),
 
     /// Indicates an error in the underlying TLS stream.
     #[cfg(feature = "tls")]
@@ -150,5 +150,5 @@ impl Protocol {
 }
 
 pub(crate) fn wrap_serde_io(err: std::io::Error) -> Error {
-    Error::DeserializePayload(serde_json::Error::io(err))
+    Error::Serialization(serde_json::Error::io(err))
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,16 +1,16 @@
 //! Enumerates all errors that this crate may return.
 //!
-//! [Error] is the top level error enum.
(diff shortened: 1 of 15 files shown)


kitified:
The codecov regression is due to the fact that we now have more explicit errors; I don't think this is an actual coverage regression. For that reason I'd like to request we overlook this.

src/error.rs
```diff
@@ -1,26 +1,65 @@
...
+/// The set of observable errors when interacting with a Faktory server.
+#[derive(Debug, Error)]
+#[allow(clippy::manual_non_exhaustive)]
+pub enum Error {
+    /// The connection to the server, or one of its prerequisites, failed.
+    #[error("connection error: {0}")]
+    Connect(#[from] ConnectError),
+
+    /// Underlying io layer errors.
+    /// These are overwhelmingly network communication errors on the socket connection to the server.
```
jonhoo, line 14: This is a nit, but in general rustdocs prefer an empty line between the first line (the "summary") and the remainder of the text.

src/error.rs
```diff
@@ -1,26 +1,65 @@
...
+    #[error("underlying i/o: {0}")]
+    GenericIO(#[from] std::io::Error),
```
jonhoo, line 16: I'm surprised this doesn't generate a warning. Rust tends to be pretty strict about only using capital letters for each word, which tends to get tripped up on things like IO. But if this doesn't generate a warning any more then :+1:

jonhoo: Any particular reason to call this `GenericIO` as opposed to just `IO`? What makes it "generic"?

src/error.rs
```diff
@@ -1,26 +1,65 @@
+use thiserror::Error;
+
+use crate::proto::ConnectError;
+
+/// The set of observable errors when interacting with a Faktory server.
+#[derive(Debug, Error)]
+#[allow(clippy::manual_non_exhaustive)]
+pub enum Error {
+    /// The connection to the server, or one of its prerequisites, failed.
+    #[error("connection error: {0}")]
```
jonhoo, line 10: I think the recommendation is to _either_ return inner errors through `Error::source` (which `#[from]` does) _or_ print the inner error in `Display`, but never both (see https://github.com/rust-lang/project-error-handling/issues/27#issuecomment-763950178, https://github.com/rust-lang/project-error-handling/issues/23, https://github.com/rust-lang/api-guidelines/pull/210). So I think these errors shouldn't include `: {0}` where `.0` is an inner error.

src/error.rs
```diff
@@ -1,26 +1,65 @@
...
+    #[error("underlying tls stream: {0}")]
+    TlsStream(#[from] native_tls::Error),
+
+    // We're going to add more error types in the future
+    // https://github.com/rust-lang/rust/issues/44109
+    //
+    // This forces users to write pattern matches with a catch-all `_` arm.
+    #[error("unreachable")]
+    #[doc(hidden)]
+    __Nonexhaustive,
```
jonhoo, line 40: I think we can safely adopt `#[non_exhaustive]` now, unless you think there's a strong reason not to?

src/error.rs
```diff
@@ -47,30 +83,32 @@ pub enum FaktoryError {
 
     // We're going to add more error types in the future
     // https://github.com/rust-lang/rust/issues/44109
-    #[fail(display = "unreachable")]
+    //
+    // This forces users to write pattern matches with a catch-all `_` arm.
+    #[error("unreachable")]
```
jonhoo, line 88: Same here — this can now use `#[non_exhaustive]`.

src/producer/mod.rs
```diff
@@ -93,38 +96,37 @@ impl Producer<TcpStream> {
 
 impl<S: Read + Write> Producer<S> {
     /// Connect to a Faktory server with a non-standard stream.
-    pub fn connect_with(stream: S, pwd: Option<String>) -> Result<Producer<S>, Error> {
-        Ok(Producer {
-            c: Client::new_producer(stream, pwd)?,
-        })
+    pub fn connect_with(stream: S, pwd: Option<String>) -> Result<Producer<S>> {
+        let c = Client::new_producer(stream, pwd)?;
+        Ok(Producer { c })
```
jonhoo, line 101: Why change the body structure here?

src/error.rs
```diff
@@ -1,26 +1,65 @@
+use thiserror::Error;
+
+use crate::proto::ConnectError;
```
jonhoo, line 3: I think we should just bring `ConnectError` into here too while we're at it.

src/lib.rs
```diff
@@ -73,7 +71,7 @@ mod tls;
 pub use tls::TlsStream;
 
 pub use crate::consumer::{Consumer, ConsumerBuilder};
-pub use crate::error::FaktoryError;
+pub use crate::error::{Error, ProtocolError};
```
jonhoo, line 74: I'm tempted to only `pub use Error`, and then make `mod error` `pub` for access to "deeper" error types.

jonhoo: If we do that, I think we should also remove `Error` from `ProtocolError` to avoid stuttering (`error::ProtocolError` -> `error::Protocol`). This is a topic under ongoing debate (https://github.com/rust-lang/api-guidelines/discussions/66), but I think that's the way things are generally leaning.

src/proto/mod.rs
```diff
@@ -17,21 +18,23 @@ pub use self::single::{Ack, Fail, Heartbeat, Info, Job, Push, QueueAction, Queue
...
-    #[fail(display = "server requires authentication")]
+    #[error("server requires authentication")]
     AuthenticationNeeded,
-    #[fail(
-        display = "server version mismatch (theirs: {}, ours: {})",
-        theirs, ours
-    )]
+    #[error("server version mismatch (theirs: {theirs}, ours: {ours})")]
     VersionMismatch { ours: usize, theirs: usize },
+    #[error("parse URL: {0}")]
```
jonhoo, line 31: Same here re `#[from]` + `{0}`.

src/proto/mod.rs
```diff
@@ -199,6 +202,7 @@ impl<S: Read + Write> Client<S> {
                 return Err(ConnectError::AuthenticationNeeded.into());
             }
         }
+
```
jonhoo, line 205: Was this intentional?

src/proto/single/cmd.rs
```diff
@@ -1,14 +1,16 @@
-use super::Job;
-use failure::Error;
+use super::{Error, Job};
 use std::io::prelude::*;
 
+/// Default Error type, aliased in module for convenience.
+type Result<T> = std::result::Result<T, Error>;
```
jonhoo, line 5: I'm not sure I see much value in this type alias. It increases the size of the diff, and means devs now need to look up the alias if editing a given function. It also means we sometimes have to qualify `std::result::Result` below, which isn't ideal.

jonhoo: (same for other files)

src/proto/single/mod.rs
```diff
@@ -71,7 +72,7 @@ pub struct Job {
     ///
     /// This field is read-only.
     #[serde(skip_serializing)]
-    pub(crate) failure: Option<Failure>,
+    pub failure: Option<Failure>,
```
jonhoo, line 75: I think I made this non-`pub` because it's read-only. It should instead have a getter function that yields `Option<&Failure>` so that users aren't tempted to try to modify it for an outgoing `Job`.

jonhoo: Also, I just realized, `Failure`'s fields have no public fields and no accessors, so even if a user _does_ get a `&Failure` there's not much they can do with it. I'd be open to making the fields of `Failure` `pub` though (although with `#[non_exhaustive]`.

src/proto/single/resp.rs
```diff
@@ -1,8 +1,9 @@
-use crate::FaktoryError;
-use failure::Error;
+use crate::ProtocolError;
 use std::io::prelude::*;
 
-fn bad(expected: &'static str, got: &RawResponse) -> FaktoryError {
+use super::Error;
```
jonhoo, line 4: I don't think you need the empty line between this and the preceding imports.

jonhoo (changes requested):
Thanks for taking this on! These are a great set of changes. I've left some mostly minor notes inline. And yes, happy to overlook the coverage change.

src/error.rs
kitified: No real reason, happy to change it back!

src/producer/mod.rs
kitified: Oh, I think that was left over from something I was doing earlier, changing it back.

src/lib.rs
kitified: Big fan of this!

src/proto/single/mod.rs
kitified: Makes sense. I'll just revert this for now and potentially revisit in another PR later!

src/proto/mod.rs
kitified: Yes, I thought having a blank line here like other blocks in this function/crate made sense. Happy to change it back if you prefer.

kitified:
@jonhoo thanks for the great review! I think I addressed everything 😄

src/proto/single/mod.rs
```diff
@@ -156,6 +156,11 @@ impl Job {
     pub fn args(&self) -> &[serde_json::Value] {
         &self.args
     }
+
+    /// Data about this job's most recent failure.
+    pub fn failure(&self) -> &Option<Failure> {
+        &self.failure
+    }
```
kitified, line 163: I think this is the minimal set of changes I wanted to include in this PR, and only did when forced to by clippy, so `Failure` is still unusable. I'm happy to commit to coming back in another PR to make `Failure` useful, or put it into this one if preferred.

src/proto/mod.rs
```diff
@@ -1,6 +1,5 @@
-use crate::FaktoryError;
+use crate::error::{Connect, Error, Protocol};
```
jonhoo, line 1: nit: my personal preference here would be to just `use crate::error::{self, Error}`, and then below write `error::Connect` to make it a little clearer at a glance. Not a blocking concern though!

src/proto/single/mod.rs
jonhoo: Let's do that in a separate PR then, thanks :+1:

src/proto/single/resp.rs
```diff
@@ -1,8 +1,7 @@
-use crate::FaktoryError;
-use failure::Error;
+use crate::error::{Error, Protocol};
```
jonhoo, line 1: Same nit here: I'd import `crate::error::{self, Error}` and use `error::Protocol` below.

src/error.rs
```diff
@@ -1,26 +1,107 @@
+//! Enumerates all errors that this crate may return.
+//!
+//! [Error] is the top level error enum.
```
jonhoo, line 3: The convention for references to code identifiers is to make them be in code-font:
```suggestion
//! [`Error`] is the top level error enum.
```

src/error.rs
```diff
@@ -1,26 +1,107 @@
...
+//! Most consumers should only need to interact with this type.
+//! This is also where more generic errors such as IO errors are placed,
```
jonhoo, line 5:
```suggestion
//! This is also where more generic errors such as I/O errors are placed,
```

src/error.rs
```diff
@@ -1,26 +1,107 @@
...
+    ///
+    /// This error is one that was encountered when attempting to deserialize a response from the server.
+    /// These generally indicate a mismatch between what the client expects and what the server provided.
+    #[error("deserialize payload")]
+    DeserializePayload(#[from] serde_json::Error),
+
+    /// Indicates an error in the underlying TLS stream.
+    #[cfg(feature = "tls")]
+    #[error("underlying tls stream")]
+    TlsStream(#[from] native_tls::Error),
```
jonhoo, line 48: Same as above, let's make this just `#[source]`.

src/error.rs
```diff
@@ -1,26 +1,107 @@
...
+    /// These generally indicate a mismatch between what the client expects and what the server expects.
+    #[error("protocol")]
+    Protocol(#[from] Protocol),
+
+    /// Faktory payloads are JSON encoded.
+    ///
+    /// This error is one that was encountered when attempting to deserialize a response from the server.
+    /// These generally indicate a mismatch between what the client expects and what the server provided.
+    #[error("deserialize payload")]
+    DeserializePayload(#[from] serde_json::Error),
```
jonhoo, line 43: I think we should make this `#[source]` — it's not clear to me that we want to commit to always having a `From` impl for `serde_json::Error`. It will mean that we have to `.map_err(Error::DeserializePayload)` in one place or another, but I think that's okay.

jonhoo (changes requested):
We're very nearly there!

src/error.rs
kitified: This required a fair amount of mapping (leading to the creation of `error::wrap_serde_io`), but I think was definitely a good change if for no other reason than that it highlighted that this error shouldn't imply it's only _de_serialization!

src/proto/single/cmd.rs
```diff
@@ -42,9 +43,9 @@ pub struct Ack {
 
 impl FaktoryCommand for Ack {
     fn issue<W: Write>(&self, w: &mut dyn Write) -> Result<(), Error> {
-        w.write_all(b"ACK ").map_err(serde_json::Error::io)?;
-        serde_json::to_writer(&mut *w, self)?;
-        Ok(w.write_all(b"\r\n").map_err(serde_json::Error::io)?)
+        w.write_all(b"ACK ").map_err(error::wrap_serde_io)?;
+        serde_json::to_writer(&mut *w, self).map_err(Error::Serialization)?;
+        w.write_all(b"\r\n").map_err(error::wrap_serde_io)
```
kitified, line 48: Do you think that these `w.write_all` should actually be `Error::Serialization`, or should they be `Error::IO`? I made them the former for consistency with the previous results, but I think they should really be the latter since it's more correct and we're breaking anyway.

src/proto/single/resp.rs
```diff
@@ -154,22 +157,24 @@ fn read<R: BufRead>(mut r: R) -> Result<RawResponse, Error> {
         }
         b'$' => {
             // Bulk String
-            // https://redis.io/topics/protocol#resp-bulk-strings
+            // https://redis.io/topics/error::protocol#resp-bulk-strings
```
jonhoo, line 160: This looks like it may have been unintentional?

src/proto/single/cmd.rs
jonhoo: I like the idea of making the `write_all` calls `Error::IO` now!

jonhoo (changes requested):
Just a few minor nits left and then we're there :tada:

src/proto/single/resp.rs
kitified: Argh, good catch 💀

jonhoo (approved):

jonhoo:
Thanks for sticking with this! I'll release it as a `.beta` initially, and then in a week or two we can promote to a full release — how does that sound?

jonhoo:
Released as 0.12.0-beta.1 :tada:

kitified:
> Thanks for sticking with this! I'll release it as a `.beta` initially, and then in a week or two we can promote to a full release — how does that sound?

Sounds great!

jonhoo:
Published 0.12.0-beta.1 as 0.12.0!

PR #29, opened 2022-08-16
Remove chrono default features

scruffystuffs:
While trying to track down why our use of chrono (which does not require any WASM) was importing WASM deps, I noticed that faktory is requiring the default features of chrono without using them.  This PR removes all unused features from chrono.

If any testing other than `cargo test` is needed, please let me know.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/29)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ sha2 = "0.10.0"
 hostname = "0.3"
 libc = "0.2"
 rand = "0.8"
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4", features = ["serde", "clock"], default-features = false }
 url = "2"
 atomic-option = "0.1"
 fnv = "1.0.3"
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,7 @@
 jobs:
  - template: default.yml@templates
    parameters:
-     minrust: 1.56
+     minrust: 1.57
  - job: integration
    displayName: cargo test (real)
    pool:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,7 +1,7 @@
 jobs:
  - template: default.yml@templates
    parameters:
-     minrust: 1.54
+     minrust: 1.56
  - job: integration
    displayName: cargo test (real)
    pool:
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -145,7 +145,7 @@ fn read<R: BufRead>(mut r: R) -> Result<RawResponse, Error> {
             let l = s.len() - 2;
             s.truncate(l);
 
-            match (&*s).parse::<isize>() {
+            match (*s).parse::<isize>() {
                 Ok(i) => Ok(RawResponse::Number(i)),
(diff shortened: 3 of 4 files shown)


scruffystuffs:
Hmm, seems `iana-time-zone` also depends on `js-sys` and other WASM stuff...

jonhoo:
This is a great change, thanks!

I'm okay with bumping the MSRV — you can change it here:

https://github.com/jonhoo/faktory-rs/blob/1054a13e2636a41fba166761d0126736e4d0ac14/azure-pipelines.yml#L4

scruffystuffs:
There's still a failing test, but it's flaky.  Running `faktory` locally in a container, I was able to reproduce the issue once, but then not while debugging, and not again after.  Notably, it only seems to happen on a fresh container.  I am able to repro consistently with the following script:

```bash
docker run --rm --name faktory-test -d -p 7419:7419 -p 7420:7420 -- contribsys/faktory
export FAKTORY_URL=tcp://localhost:7419
cargo test
docker container stop faktory-test
```

Changing the `cargo test` line to `cargo test || cargo test` (i.e. running it twice) consistently _succeeds_.
Waiting up to 1 minute between starting the container and running `cargo test` has no effect.

I can dig into this, but some helping ideas would be great.

jonhoo:
That's interesting. Does the test also fail on `master`, or just with this change? I wonder if there's a race here, and simply adding a `std::thread::sleep(Duration::from_millis(500))` between the `enqueue` and the `run_one` would be sufficient:
https://github.com/jonhoo/faktory-rs/blob/1054a13e2636a41fba166761d0126736e4d0ac14/tests/real.rs#L153

scruffystuffs:
(_note: while writing this comment I might have solved it_)

No, this fails on `master` as well.  I've tried sleeping in various places, including between `docker run ... faktory` and `cargo test`, and it seems to have no effect.  I don't think this is a race, I think it may be something to do with either a conflict or some idempotent pre-condition.

For context, the relevant test code is this:
```rust
    let mut p = Producer::connect(None).unwrap();
    p.queue_pause(&[local]).unwrap();
    p.enqueue(Job::new(local, vec![Value::from(1)]).on_queue(local))
        .unwrap();


    let had_job = c.run_one(0, &[local]).unwrap();
    assert!(!had_job);
```

My guess is that on a fresh container, `queue_pause` is a no-op, because the queue doesn't exist yet.  Then, when `enqueue` is called, the queue is created (and not paused), the job exists, and the expectation (that the job won't be run because the queue is paused) fails.

I was able to successfully run the test by switching the order of `queue_pause` and `enqueue`.

jonhoo (approved):
Huh, that's interesting. It seems wrong to be unable to pause a queue before enqueueing anything on it, but I don't mind changing the test this way to work around it. May be worth reporting upstream!

jonhoo:
Released as 0.12.1 :tada:

scruffystuffs:
I think that it's because the queue just doesn't exist yet.  If we were to somehow create the queue with no items in it, pause it, and then enqueue, we could replicate the old behavior of the test exactly.

I can report upstream that pausing a queue no longer creates the queue if it doesn't exist.

PR #32, opened 2023-08-06
Minor: Clap was outdated

martinfrances107:
-clap = { version = "3.1.0", optional = true }
+clap = { version = "4.3.19", optional = true }

Also ran cargo outdated and updated many sub-packages.

As per comment on "Impromptu stream: open-source maintenance" ..  could use some TLC.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/32)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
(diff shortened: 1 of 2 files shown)


jonhoo:
Upgrading clap 3 to clap 4 is, unfortunately, [a bit more work](https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating) to ensure that everything works the same. If you're willing to go through those steps though, that'd be great!

You shouldn't need to update anything else beyond the one clap version number in `Cargo.toml` , and then run `cargo test` which will update only required dependencies in `Cargo.lock`. Unless there is specifically a reason to update the `Cargo.lock`, I'd prefer to not proactively update it more so than is necessary as part of this PR.

PR #36, opened 2023-12-02
Cover loadtest bin with tests

rustworthy:
- Refactoring loadtest binary to make it testable;
- Adding tests for loadtest bin to increase overall coverage;

Loadtest functions as expected:
image

Coverage report for loadtest.rs (with suggested changes applied):
image

Overall ~10p.p. coverage increase from 69.88% to 79.46%:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/36)

### DIFF
--- a/src/bin/loadtest.rs
+++ b/src/bin/loadtest.rs
@@ -152,19 +152,16 @@ fn calc_secs_elapsed(elapsed: &time::Duration) -> f64 {
     elapsed_nanos as f64 / 1_000_000_000.0
 }
 
-fn main() {
-    let opts = get_opts(None);
-    let jobs_count = opts.get("jobs").unwrap().to_owned();
-    let threads_count = opts.get("threads").unwrap().to_owned();
+fn run_loadtest(jobs: usize, threads: usize) {
     println!(
         "Running loadtest with {} jobs and {} threads",
-        jobs_count, threads_count
+        jobs, threads
     );
 
     ping!();
 
     let start = time::Instant::now();
-    let (_ops_count, pushed, popped) = load_with_jobs(jobs_count, threads_count);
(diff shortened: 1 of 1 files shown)


jonhoo:
I'm not sure I follow the reasoning behind this change. `loadtest` is intended to be a simple program for running load testing against Faktory, and I don't think it's particularly important that it itself be tested (beyond the fact that it compiles). I'd be more inclined to simply exclude it from the coverage calculation if we're worried about it artificially dragging the number down.

rustworthy:
> I'm not sure I follow the reasoning behind this change. `loadtest` is intended to be a simple program for running load testing against Faktory, and I don't think it's particularly important that it itself be tested (beyond the fact that it compiles). I'd be more inclined to simply exclude it from the coverage calculation if we're worried about it artificially dragging the number down.

I believe it's easier to read and reason about when it's decomposed. Say, we will want to load test with some of new features in place: job tracker, who will send a job execution updates once the job is fetched. Then it will be easier to adjust the binary to accept these options from the command line. The fact (fakt) that we can now cover these separate procedures with tests is rather a side-effect, isn't it? :)

jonhoo:
Hmm, I actually find the [new version](https://github.com/jonhoo/faktory-rs/blob/b695db42fcd35c222ad59dbbc0bee8191db19ca3/src/bin/loadtest.rs) significantly harder to read/follow than the [old version](https://github.com/jonhoo/faktory-rs/blob/481b15241eff2e138695e19f9af998c3caffbad4/src/bin/loadtest.rs) :sweat_smile:  I'd prefer to keep the old one until such time as we need the added flexibility of the more convoluted new one :)

Issue #9, opened 2018-03-18
Use builder for creating Job


wezm:
Currently setting properties on Job outside of kind and args requires manually building a Job or building a Job with new, then mutating its properties. I'd suggest using a builder to create jobs and make the fields of Job private. My suggested approach would probably be [derive_builder](https://crates.io/crates/derive_builder).

If this sounds reasonable I'm happy to make this change.

jonhoo:
Yeah, that makes a lot of sense! I've been wanting to make that change in accordance with [`C-BUILDER`](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#c-builder) in the Rust API guidelines too. A PR using `derive_builder` seems reasonable to me!

jonhoo:
@wezm still think you'd like to make this change? I'd be happy to review and merge a change like this!

rustworthy:
Implemented with the help of `derive_builder` crate and released in `v0.13.0`

PR #34, opened 2023-11-30
Bump clap from 3.2.25 to 4.4.12 and MSRV from 1.60.0 to 1.70.0

rustworthy:
Move from clap@3 to clap@4 and set msrv to 1.70

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/34)

PR #37, opened 2023-12-03
 Job builder

rustworthy:
As per [feature request](https://github.com/jonhoo/faktory-rs/issues/9)

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/37)

### DIFF
--- a/src/bin/loadtest.rs
+++ b/src/bin/loadtest.rs
@@ -70,26 +70,12 @@ fn main() {
                 for idx in 0..jobs {
                     if idx % 2 == 0 {
                         // push
-                        let job = match idx % 4 {
-                            // Solely to demo how 'JobBuilder' works in userland
-                            0 => JobBuilder::default()
-                                .kind("SomeJob")
-                                .args(vec![serde_json::Value::from(1), "string".into(), 3.into()])
-                                .priority(Some(rng.gen_range(1..10)))
-                                .queue(QUEUES.choose(&mut rng).unwrap().to_string())
-                                .build()
-                                .unwrap(),
-                            _ => {
-                                let mut job = Job::new(
-                                    "SomeJob",
-                                    vec![serde_json::Value::from(1), "string".into(), 3.into()],
-                                );
-                                job.priority = Some(rng.gen_range(1..10));
-                                job.queue = QUEUES.choose(&mut rng).unwrap().to_string();
-                                job
-                            }
-                        };
-
+                        let mut job = Job::new(
+                            "SomeJob",
+                            vec![serde_json::Value::from(1), "string".into(), 3.into()],
+                        );
+                        job.priority = Some(rng.gen_range(1..10));
+                        job.queue = QUEUES.choose(&mut rng).unwrap().to_string();
                         p.enqueue(job)?;
                         if pushed.fetch_add(1, atomic::Ordering::SeqCst) >= jobs {
                             return Ok(idx);
--- a/src/bin/loadtest.rs
+++ b/src/bin/loadtest.rs
@@ -70,12 +70,26 @@ fn main() {
                 for idx in 0..jobs {
                     if idx % 2 == 0 {
                         // push
-                        let mut job = Job::new(
-                            "SomeJob",
-                            vec![serde_json::Value::from(1), "string".into(), 3.into()],
-                        );
-                        job.priority = Some(rng.gen_range(1..10));
-                        job.queue = QUEUES.choose(&mut rng).unwrap().to_string();
+                        let job = match idx % 4 {
+                            // Solely to demo how 'JobBuilder' works in userland
+                            0 => JobBuilder::default()
+                                .kind("SomeJob")
+                                .args(vec![serde_json::Value::from(1), "string".into(), 3.into()])
+                                .priority(Some(rng.gen_range(1..10)))
+                                .queue(QUEUES.choose(&mut rng).unwrap().to_string())
+                                .build()
+                                .unwrap(),
+                            _ => {
+                                let mut job = Job::new(
+                                    "SomeJob",
+                                    vec![serde_json::Value::from(1), "string".into(), 3.into()],
+                                );
+                                job.priority = Some(rng.gen_range(1..10));
+                                job.queue = QUEUES.choose(&mut rng).unwrap().to_string();
+                                job
+                            }
+                        };
+
                         p.enqueue(job)?;
                         if pushed.fetch_add(1, atomic::Ordering::SeqCst) >= jobs {
                             return Ok(idx);
--- a/src/error.rs
+++ b/src/error.rs
@@ -25,11 +25,13 @@ pub enum Error {
     /// The connection to the server, or one of its prerequisites, failed.
     #[error("connection")]
     Connect(#[from] Connect),
+
     /// Client-side errors.
     ///
     /// These are errors arising even before submitting a job to the server, e.g. malformed job.
     #[error("client")]
     Client(#[from] Client),
+
     /// Underlying I/O layer errors.
     ///
     /// These are overwhelmingly network communication errors on the socket connection to the server.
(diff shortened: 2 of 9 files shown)


src/bin/loadtest.rs
```diff
@@ -70,12 +70,26 @@ fn main() {
...
                     if idx % 2 == 0 {
                         // push
-                        let mut job = Job::new(
-                            "SomeJob",
-                            vec![serde_json::Value::from(1), "string".into(), 3.into()],
-                        );
-                        job.priority = Some(rng.gen_range(1..10));
-                        job.queue = QUEUES.choose(&mut rng).unwrap().to_string();
+                        let job = match idx % 4 {
+                            // Solely to demo how 'JobBuilder' works in userland
```
jonhoo, line 74: This feels like it belongs in a code block in the documentation. The `loadtest` binary isn't really intended as user-facing documentation, but rather as a stand-lone tool.

src/error.rs
```diff
@@ -22,7 +25,11 @@ pub enum Error {
     /// The connection to the server, or one of its prerequisites, failed.
     #[error("connection")]
     Connect(#[from] Connect),
-
```
jonhoo, line 25: nit: please preserve the empty line between each enum variant

src/proto/single/mod.rs
```diff
@@ -83,10 +128,41 @@ pub struct Job {
...
+        let job = self
+            .try_build()
+            .map_err(|err| error::Client::MalformedJob {
+                desc: err.to_string(),
+            })?;
+        Ok(job)
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
```
jonhoo, line 165: This should probably continue to not be `PartialEq`.

src/error.rs
```diff
@@ -82,6 +89,18 @@ pub enum Connect {
     ParseUrl(#[source] url::ParseError),
 }
 
+/// Errors happening client side
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum Client {
+    /// The 'Job' is malformed.
+    #[error("job is malformed: {desc}")]
+    MalformedJob {
```
jonhoo, line 98: Do we ever expect to have other client-side errors than malformed jobs? I wonder if we should skip the intermediate `enum` here and instead have `MalformedJob` be a variant directly on `Error`.

src/proto/single/mod.rs
```diff
@@ -1,77 +1,122 @@
...
+/// use faktory::JobBuilder;
+///
+/// let result = JobBuilder::default()
+///     .kind("order")
+///     .args(vec!["ISBN-13:9781718501850"])
+///     .build();
+/// if result.is_err() {
+///     todo!("Handle me gracefully, please.")
+/// };
+/// let _job = result.unwrap();
```
jonhoo, line 42:
```suggestion
/// let _job = JobBuilder::default()
///     .kind("order")
///     .args(vec!["ISBN-13:9781718501850"])
///     .build()?;
/// # Ok(())
```

src/proto/single/mod.rs
```diff
@@ -1,77 +1,122 @@
...
+    #[builder(default = "utils::gen_random_jid()")]
     pub(crate) jid: String,
 
     /// The queue this job belongs to. Usually `default`.
+    #[builder(default = "JOB_DEFAULT_QUEUE.into()")]
     pub queue: String,
 
     /// The job's type. Called `kind` because `type` is reserved.
     #[serde(rename = "jobtype")]
     pub(crate) kind: String,
```
jonhoo, line 62: Given that `kind` is a required argument, I'd rather it have to be passed to `builder`. That seems to be the generally-recommended pattern for options that must be set for the build to succeed. Look at the doc code elsewhere, that doesn't seem to be the case at the moment?

src/proto/single/mod.rs
```diff
@@ -108,26 +184,19 @@ impl Job {
...
         A: Into<serde_json::Value>,
     {
-        use rand::{thread_rng, Rng};
-        let random_jid = thread_rng()
-            .sample_iter(&rand::distributions::Alphanumeric)
-            .map(char::from)
-            .take(16)
-            .collect();
-        use chrono::prelude::*;
         Job {
```
jonhoo, line 187: To avoid duplicating code here, can we now make this just call `JobBuilder::new(kind).args(args).build()`?

src/proto/single/mod.rs
jonhoo: An alternative if the derive crate doesn't let you do that is to have it generate a private constructor (i.e., to make `JobBuilder` not implement `Default` ), and instead have a `Job::builder(kind)` that returns a `JobBuilder` where the `kind` has been set.

src/proto/single/mod.rs
```diff
@@ -83,10 +128,41 @@ pub struct Job {
...
+        self.args = Some(args.into_iter().map(|s| s.into()).collect());
+        self
+    }
+
+    fn validate(&self) -> Result<(), String> {
+        if let Some(ref priority) = self.priority {
+            if *priority > Some(JOB_PRIORITY_MAX) {
+                return Err("`priority` must be in the range from 0 to 9 inclusive".to_string());
+            }
+        }
```
jonhoo, line 150: If this is the only kind of misconfiguration there can be for a job, I'm inclined to not have it return an error (and thus get rid of the new error type and the `Result` in `build`'s return type. Especially because `priority` is already a `pub` field, so users can always set it to the wrong value elsewhere.

jonhoo (changes requested):
This looks like a good starting point, thanks for taking the time! Left some comments.

src/proto/single/mod.rs
rustworthy: This 'default' in current implementation is definitely deceptive: as if it were enough to build a simple job, while it's not and it really should be followed/enriched by `.kind()`  and `.args()` , otherwise the error will be returned.

With these updates:

1)  as per your suggestion, the required arg is set of the JobBuilder right away (either via JobBuilder::new("jobtype" or  - also added as an alternative way to instantiate a builder - via Job::builder("jobtype")));

2) the 'args' field has been given a default of 'Vec::new()' on the JobBuilder;

3) as per your suggestions, we got rid of the 'priority' validation (As a side-note: on the second thought, I realized that adding a 'JobPriority' (e.g. JobPriority::from(3)) new type would have been a better solution compared to erroring back at a job's build time. In the new type case they would be able to set/reset the priority as many time as they want and it would still be valid. But this 1) looks like a subject of a different PR and 2) not necessary in the first place);

We can:
1. Create a new Job with JobBuilder infallibly;

2. Get rid of the 'Client' error variant (I am using this one in another branch on tracking job's progress for Faktory enterprise, BUT I am using it there in the exact same case: checking that 'percent' value in the range from 0 to 100. I now think I should be adding this validation there to not constrain the user and plus no instructions on this are given in Faktory docs).

src/proto/single/mod.rs
rustworthy: Got rid of this.

src/proto/single/mod.rs
rustworthy: Added usage examples taking into account that the build is not returning result any longer

src/error.rs
rustworthy: Looks like with other suggestions of yours applied, this variant is no longer needed (for now :))

src/bin/loadtest.rs
rustworthy: restored

jonhoo (changes requested):

jonhoo (approved):

jonhoo:
I'll hold off on releasing this until #34 lands.

jonhoo:
Released in 0.12.2 :tada:

PR #45, opened 2024-01-03
Unique and expiring jobs (Enterprise Faktory)

rustworthy:
Putting those features under the feature called "ent", because looks like it is already a common abbreviation/label for Enterprise Faktory.

Referencing the Faktory's [wiki](https://github.com/contribsys/faktory/wiki) with regard to API.

Covering these Ent Faktory features:
- [expiring jobs](https://github.com/contribsys/faktory/wiki/Ent-Expiring-Jobs)
- [unique jobs](https://github.com/contribsys/faktory/wiki/Ent-Unique-Jobs)

---

Will be added as separate PRs:
- [batches](https://github.com/contribsys/faktory/wiki/Ent-Batches)
- [tracking](https://github.com/contribsys/faktory/wiki/Ent-Tracking)

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/45)


.github/workflows/ent.yaml
```diff
@@ -0,0 +1,47 @@
...
+          FAKTORY_ENT: true
+        run: cargo test --locked --features ent --all-targets
+  doc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install stable
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run doc tests
+        run: cargo test --features ent --doc
```
jonhoo, line 47: You don't need this one since we already test + check with `--all-features` elsewhere.

.github/workflows/ent.yaml
```diff
@@ -0,0 +1,47 @@
+# This is a CI workflow that runs the test against Enterprise Edition of Faktory.
+# The binary (for macos only) is avalable for download for testing purposes with each Faktory release.
```
jonhoo, line 2: Nice!

src/proto/single/ent.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.add_to_custom_data(
+            "expires_at".into(),
+            dt.to_rfc3339_opts(chrono::SecondsFormat::Nanos, true),
+        )
+    }
+
+    /// In what period of time from now (UTC) the Faktory should expire this job.
+    ///
+    /// Use this setter when you are unwilling to populate the `expires_at` field in custom
+    /// options with some exact date and time, e.g.:
```
jonhoo, line 29: Given the implementation of this method is to compute `Utc::now` and then set an absolute time, this comment is a little misleading — a caller might reasonably read it and conclude that they can use this to use the server's time or something like that. May be worth clarifying that this just computes `now + ttl` under the hood.

src/proto/single/ent.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.expires_at(Utc::now() + ttl)
+    }
+
+    /// How long the Faktory will not accept duplicates of this job.
+    ///
+    /// The job will be considered unique for kind-args-queue combination.
+    /// The uniqueness is best-effort, rather than a guarantee. Check out
+    /// the Enterprise Faktory [docs](https://github.com/contribsys/faktory/wiki/Ent-Unique-Jobs)
+    /// for details on how scheduling, retries and other features live together with `unique_for`.
+    pub fn unique_for(&mut self, secs: usize) -> &mut Self {
```
jonhoo, line 49: I think commonly `u64` is used for seconds in this setting (e.g., `std::time::Duration::as_secs`).

src/proto/single/mod.rs
```diff
@@ -162,6 +165,13 @@ impl JobBuilder {
         self
     }
 
+    /// Sets arbitrary key-value pairs to this job's custom data hash.
+    pub fn add_to_custom_data(&mut self, k: String, v: impl Into<serde_json::Value>) -> &mut Self {
```
jonhoo, line 169: I'd probably make `k: impl Into<String>` as well to avoid all the `.into()`s in the calls to this function.

tests/real/enterprise.rs
```diff
@@ -0,0 +1,321 @@
...
+        .expires_at(chrono::Utc::now() + ttl)
+        .build();
+
+    // enquere and then fetch job2, but after ttl:
+    producer.enqueue(job2).unwrap();
+    thread::sleep(time::Duration::from_secs(job_ttl_secs * 2));
+    let had_job = consumer.run_one(0, &["default"]).unwrap();
+
+    // For the non-enterprise edition of Faktory, this assertion will
+    // fail, which should be take into account when running the test suite on CI.
```
jonhoo, line 69:
```suggestion
    // fail, which should be taken into account when running the test suite on CI.
```

tests/real/enterprise.rs
```diff
@@ -0,0 +1,321 @@
...
+    // this one is a 'duplicate' ...
+    let job2 = Job::builder(job_type)
+        .args(args.clone())
+        .queue(queue_name)
+        .unique_for(unique_for_secs)
+        .build();
+    // ... so the server will respond accordingly:
+    let res = producer.enqueue(job2).unwrap_err();
+    if let error::Error::Protocol(error::Protocol::Internal { msg }) = res {
+        assert_eq!(msg, "NOTUNIQUE Job not unique");
```
jonhoo, line 134: Feels worthwhile to make this a dedicated error variant I think.

tests/real/enterprise.rs
```diff
@@ -0,0 +1,321 @@
...
+    } else {
+        panic!("Expected protocol error.")
+    }
+
+    let had_job = consumer.run_one(0, &[queue_name]).unwrap();
+    assert!(had_job);
+    let had_another_one = consumer.run_one(0, &[queue_name]).unwrap();
+
+    // For the non-enterprise edition of Faktory, this assertion WILL FAIL:
```
jonhoo, line 143: Won't it hit the `panic!` above first? Or, indeed, the `.unwrap_err`.

tests/real/enterprise.rs
```diff
@@ -0,0 +1,321 @@
...
+        assert_eq!(msg, "NOTUNIQUE Job not unique");
+    } else {
+        panic!("Expected protocol error.")
+    }
+
+    handle.join().expect("should join successfully");
+
+    // Not that the job submitted in a spawned thread has been successfully executed
+    // (with ACK sent to server), the producer 'B' can push another one:
+    assert!(producer_b
```
jonhoo, line 248: Instead of `assert!(...is_ok())`, you can just `.unwrap()` it instead.

tests/real/enterprise.rs
```diff
@@ -0,0 +1,321 @@
...
+                serde_json::from_value::<i64>(diffuculty_level).expect("a valid number");
+            thread::sleep(time::Duration::from_secs(sleep_secs as u64));
+            Ok(eprintln!("{:?}", job))
+        });
+        let mut consumer_a = consumer_a.connect(Some(&url1)).unwrap();
+        let job = JobBuilder::new(job_type)
+            .args(vec![difficulty_level])
+            .queue(queue_name)
+            .unique_for(unique_for)
+            .unique_until_success() // Faktory's default
```
jonhoo, line 215: What happens when these are both set? Does it change the behavior in a particular way? Would be worth adding to the docs (and same with `unique_until_start`).

jonhoo (changes requested):
This looks great, thanks for taking it on!

Only a few smaller notes.

src/proto/single/ent.rs
rustworthy: This is to keep it consistent with`reserve_for` on the `Job` struct, which is an option of usize. 

Also - I've  just checked out of  curiosity - they use _uint_ in the Go [client](https://github.com/contribsys/faktory/blob/main/client/job.go#L103).

As for duration is seconds generally, in `chrono` the [`seconds`](https://docs.rs/chrono/latest/chrono/struct.Duration.html#method.seconds) method will also expect a 64bit integer, but  a _singed_ one, allowing for negative duration.

tests/real/enterprise.rs
rustworthy: Indeed,  there would not be any unique constraint violation, and asserting so would panic. Updated the comment to simply claim the queue is empty, showing the 'duplicate' job has not been queued indeed.

tests/real/enterprise.rs
rustworthy: Added more details to the docs for both `unique_until_start` and `unique_until_success`. Besides, I added a test for unique lock bypassing, which is something I missed on the first iteration. And since it has found its place in the Faktory [docs](https://github.com/contribsys/faktory/wiki/Ent-Unique-Jobs#bypassing-uniqueness), I added a note on this to the `unique_for` setter docs.

tests/real/enterprise.rs
rustworthy: Added `UniqueConstraintViolation` variant to the  `Protocol` enum. Or you believe this deserves a place in the top-level faktory::Error?

tests/real/enterprise.rs
jonhoo: Perfect, thank you!

tests/real/enterprise.rs
```diff
@@ -321,3 +321,43 @@ fn ent_unique_job_until_start() {
...
+        .unique_for(60) // NB
+        .build();
+
+    let res = producer.enqueue(job3).unwrap_err(); // NOT bypassing the lock!
+
+    if let error::Error::Protocol(error::Protocol::UniqueConstraintViolation { msg }) = res {
+        assert_eq!(msg, "Job not unique");
+    } else {
+        panic!("Expected protocol error.")
+    }
```
jonhoo, line 362: Should we also check that both `job1` and `job2` _were_ enqueued?

tests/real/enterprise.rs
jonhoo: No, I think `Protocol` is probably a good place for it :+1:

jonhoo (approved):
This looks good now, thank you!

One thing I'd recommend is to make use of the [`doc_cfg` nightly feature](https://github.com/rust-lang/rust/issues/43781) specifically on docs.rs to highlight things that are only available with `ent`. You can see an example of how to do this in fantoccini below. We can do that in a separate PR though!

https://github.com/jonhoo/fantoccini/blob/b99209b67b176a617751885ca241189ea9c9940c/Cargo.toml#L51-L53

https://github.com/jonhoo/fantoccini/blob/b99209b67b176a617751885ca241189ea9c9940c/.github/workflows/check.yml#L49-L61

https://github.com/jonhoo/fantoccini/blob/b99209b67b176a617751885ca241189ea9c9940c/src/lib.rs#L143

https://github.com/jonhoo/fantoccini/blob/b99209b67b176a617751885ca241189ea9c9940c/src/lib.rs#L179

jonhoo:
Released in 0.12.3 :tada:

tests/real/enterprise.rs
rustworthy: Will be addressed in [PR](https://github.com/jonhoo/faktory-rs/pull/46)

PR #47, opened 2024-01-08
Makefile with useful repetitive commands

rustworthy:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/47)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: check doc faktory faktory/* test test/*
+
 FAKTORY_HOST=127.0.0.1
 FAKTORY_PORT=7419
 FAKTORY_PORT_UI=7420
@@ -22,9 +24,6 @@ faktory:
 faktory/kill:
 	docker stop faktory
 
-fmt: 
-	cargo fmt
-
 README.md: README.tpl src/lib.rs
 	cargo readme > README.md
 
@@ -38,4 +37,13 @@ test/e2e:
 	FAKTORY_URL=tcp://${FAKTORY_HOST}:${FAKTORY_PORT} cargo test --locked --all-features --all-targets
 
 test/load:
-	 cargo run --features binaries
+	cargo run --release --features binaries
+
+test/perf:
+	cargo build --release --features binaries
+	perf record -o perf_loadtest.data --call-graph dwarf target/release/loadtest
+	perf script -i perf_loadtest.data | inferno-collapse-perf > perf_loadtest_stacks.folded
+	cat perf_loadtest_stacks.folded | inferno-flamegraph > perf_loadtest.svg
+
+test/perf/clean:
+	rm perf_loadtest*
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,38 @@
+FAKTORY_HOST=127.0.0.1
+FAKTORY_PORT=7419
+FAKTORY_PORT_UI=7420
+
+check:
+	cargo fmt --check
+	cargo clippy
+	cargo d --no-deps --all-features
+
+doc:
+	RUSTDOCFLAGS='--cfg docsrs' cargo +nightly d --all-features --open
+
+faktory:
+	docker run --rm -d \
+	-v faktory-data:/var/lib/faktory \
+	-p ${FAKTORY_HOST}:${FAKTORY_PORT}:7419 \
+	-p ${FAKTORY_HOST}:${FAKTORY_PORT_UI}:7420 \
+	--name faktory \
+	contribsys/faktory:latest \
+	/faktory -b :7419 -w :7420
+
+faktory/kill:
+	docker stop faktory
+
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,41 @@
...
+	--name faktory \
+	contribsys/faktory:latest \
+	/faktory -b :7419 -w :7420
+
+faktory/kill:
+	docker stop faktory
+
+fmt: 
+	cargo fmt
+
```
jonhoo, line 27: Let's remove this — people usually have auto-formatting turned on in the editor anyway.

Makefile
```diff
@@ -0,0 +1,41 @@
...
+readme:
```
jonhoo, line 28:
```suggestion
README.md: README.tpl src/lib.rs
```
so it gets re-executed correctly.

Makefile
```diff
@@ -0,0 +1,41 @@
...
+	cargo t --locked --all-features --all-targets
+
+test/doc:
+	cargo test --locked --all-features --doc
+
+test/e2e:
+	FAKTORY_URL=tcp://${FAKTORY_HOST}:${FAKTORY_PORT} cargo test --locked --all-features --all-targets
+
+test/load:
+	 cargo run --features binaries
```
jonhoo, line 41: This should probably include `--release`

Makefile
```diff
@@ -0,0 +1,41 @@
+FAKTORY_HOST=127.0.0.1
+FAKTORY_PORT=7419
+FAKTORY_PORT_UI=7420
+
+check:
```
jonhoo, line 5: I think this and the other targets that don't produce a file should probably be declared as [phony targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html).

jonhoo (changes requested):
I suppose this might be useful — just a few bits I'd adjust.

Makefile
rustworthy: for sure

Makefile
rustworthy: With `README.md` as a target - `.PHONY` list will be even required I guess. But even if there were no 'real' targets, explicitly listing phony targets still makes sense due to this (which I honestly never considered before):
```
The implicit rule search ... is skipped for .PHONY targets. This is why declaring a target as .PHONY is good for performance, even if you are not worried about the actual file existing.
```

Thanks for pointing at that. Fixed.

---

Please have a look at `test/perf` and `test/perf/clean`  (instructions taken from  [inferno](https://docs.rs/inferno/latest/inferno/) docs).  I am now working on adding [async](https://github.com/jonhoo/faktory-rs/pull/49) support and, preliminary, I noticed a drop in performance (20-25% less jobs processed per second), which 1) of course can be implementation issue 2) could be even worse if CPU-intensive work were performed in the load test. This is definitely not subject of this PR, but I just realized that having a flamegraph target could be useful.

Makefile
```diff
@@ -0,0 +1,49 @@
+.PHONY: check doc faktory faktory/* test test/*
```
jonhoo, line 1: I don't know if `*` actually works here. Usually, Make is very particular about what `*` means, and I don't think it can wildcard rules this way when they aren't targeting pre-existing files.

Makefile
jonhoo: Looks good to me! Could alternatively use `cargo flamegraph` here which does some of this for you, and also works on other platforms.

Makefile
rustworthy: I tested it with `README.md` example where putting `RE*` to the list of phony targets will deem `README.md` always stale

rustworthy: putting as little as 1letter with a wildcard `R*` to the list will be enough to make it recreate the recipe ...
image


while putting `*` will cover all the commands:
image

jonhoo: Thanks for checking!

rustworthy: Looked through some repos. 

Indeed, this is the guidelines  they follow in the [linux kernel](https://github.com/torvalds/linux/blob/master/Makefile#L1957-L1960) repo, though  with a little adaptation - via a variable collecting the phony targets across the file and announcing them phony at the very bottom. When the file gets long, it makes it more explicit, definitely. 

In [https://github.com/redis/redis/blob/unstable/Makefile] the file is a few lines of code, and I am not sure if  they meant to put the `.PHONY` declaration at the very bottom. In [RabbitMQ](https://github.com/rabbitmq/rabbitmq-server/blob/main/Makefile#L374-L408) they declare phony target per section, which is almost the same as per target due to how they visually separate those section.

Updated according to the guidelines and checked with command:
image

`test/e2e` is _now_ really recognized as a phony target:
image

jonhoo: Excellent, thanks!

jonhoo (approved):

Issue #18, opened 2019-01-23
Provide ergonomic access to Pro features


jonhoo:
Faktory now has a [Pro edition](https://contribsys.com/faktory/), and that comes with (at the time of writing) two new API-facing features. Specifically, it adds [unique jobs](https://github.com/contribsys/faktory/wiki/Pro-Unique-Jobs) and [expiring jobs](https://github.com/contribsys/faktory/wiki/Pro-Expiring-Jobs). It'd be good to have a convenient interface to these without having to set fields in `job.custom` manually in the right format. These additional convenient methods should probably be under a non-default feature called "pro" so that non-pro users don't see methods they can't use.

jonhoo:
@mperham In line with https://github.com/contribsys/faktory/issues/180 I would love if the official client libraries did this in their CIs so other clients could copy their approach

mperham:
Unfortunately I can't provide the Linux binary necessary for most CI environments.

jonhoo:
@mperham that's fine -- a [macOS-only CI test](https://docs.travis-ci.com/user/reference/osx) would be fine. Anything automated helps!

worst:
@jonhoo are you actively working on this? I wanted to check before I start devoting some cycles to it. Expiring and unique jobs are relatively easy to get going I think.

jonhoo:
I'm passively maintaining the project, but not actively working on this feature in particular, no. So please do feel free. I'm also looking for maintainers if anyone using Faktory on a regular basis is looking to step up to the plate :)

rustworthy:
The [JobBuilder PR](https://github.com/jonhoo/faktory-rs/pull/37) followed by [expiring and  unique jobs PR](https://github.com/rustworthy/faktory-rs/pull/1) will be the first steps in enabling ergonomic access to Enterprise Faktory features. 

API for batch and trackable jobs will be added next.

PR #48, opened 2024-01-17
Batches and trackable jobs

rustworthy:
As per [issue](https://github.com/jonhoo/faktory-rs/issues/18)

- Add support for batch jobs
- Add `tracker` responsible for tracking jobs and batch statuses;

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/48)


src/producer/mod.rs
```diff
@@ -129,6 +134,25 @@ impl<S: Read + Write> Producer<S> {
...
+    /// Initiate a new batch of jobs.
+    #[cfg(feature = "ent")]
+    pub fn start_batch(&mut self, batch: Batch) -> Result<BatchHandle<'_, S>, Error> {
+        let bid = self.c.issue(&batch)?.read_bid()?;
+        Ok(BatchHandle::new(bid, self))
+    }
+
+    /// Open an already existing batch of jobs.
+    #[cfg(feature = "ent")]
+    pub fn open_batch(&mut self, bid: String) -> Result<BatchHandle<'_, S>, Error> {
```
jonhoo, line 147: Can we make this return `Result<Option` so that if there is no batch with the given `bid` we return `Ok(None`?

src/producer/mod.rs
```diff
@@ -129,6 +134,25 @@ impl<S: Read + Write> Producer<S> {
...
+        let bid = self.c.issue(&OpenBatch::from(bid))?.read_bid()?;
```
jonhoo, line 148: Should we error if the returned `bid` is not `==` the provided `bid`?

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
+use std::io::{Read, Write};
+
+use chrono::{DateTime, Utc};
+use derive_builder::Builder;
+
+use crate::{Error, Job, Producer};
```
jonhoo, line 6: nit: I don't think the whitespace-separation adds much here.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+/// ```no_run
+/// # use faktory::Error;
+/// use faktory::{Producer, Job, Batch};
+///
+/// let mut prod = Producer::connect(None)?;
+/// let job1 = Job::builder("image").build();
+/// let job2 = Job::builder("image").build();
+/// let job_cb = Job::builder("clean_up").build();
+///
+/// let batch = Batch::builder("Image resizing workload".to_string()).with_complete_callback(job_cb);
```
jonhoo, line 37: The batch and job names here and below feel a little random. Maybe use something like "First batch" or "Outer batch"?

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// belonging to this batch have been executed, even if one/some/all
+    /// of the workers have failed.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[builder(setter(custom))]
+    pub(crate) complete: Option<Job>,
+}
+
+impl Batch {
+    /// Create a new `BatchBuilder`.
+    pub fn builder(description: impl Into<Option<String>>) -> BatchBuilder {
```
jonhoo, line 136: I think I would recommend we simply disallow name-less batches, and make this argument `impl Into<String>`. It'll make the interface nicer to use ultimately, and get rid of a bunch of `.to_string()` calls.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+/// parent.commit()?;
+///
+/// # Ok::<(), Error>(())
+/// ```
+///
+/// In the example above, there is a single level nesting, but you can nest those batches as deep as you wish,
+/// effectively building a pipeline this way, since the Faktory guarantees that callback jobs will not be queued unless
+/// the batch gets committed.
+///
+/// You can retieve the batch status using a [`tracker`](struct.Tracker.html):
```
jonhoo, line 79:
```suggestion
/// You can retieve the batch status using a [`Tracker`]:
```

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+/// let cb_job = Job::builder("clean_up").build();
+/// let b = Batch::builder("Description...".to_string()).with_complete_callback(cb_job);
+///
+/// let mut b = prod.start_batch(b)?;
+/// let bid = b.id().to_string();
+/// b.add(job)?;
+/// b.commit()?;
+///
+/// let mut t = Tracker::connect(None).unwrap();
+/// let s = t.get_batch_status(bid).unwrap().unwrap();
```
jonhoo, line 94: Why use `unwrap` instead of `?` here?

src/proto/batch/mod.rs
jonhoo: Alternatively, don't take any arguments to `builder`, and instead just have people use the builder to set the name if they so wish.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// Create a new `BatchBuilder`.
+    pub fn builder(description: impl Into<Option<String>>) -> BatchBuilder {
+        BatchBuilder::new(description)
+    }
+}
+
+impl BatchBuilder {
+    fn build(&mut self) -> Batch {
+        self.try_build()
+            .expect("All required fields have been set.")
```
jonhoo, line 144:
```suggestion
            .expect("There are no required fields.")
```

src/proto/batch/mod.rs
jonhoo: Whichever way you go, I'd do the same thing for `BatchBuilder::new` below.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+        self
+    }
+
+    fn complete(&mut self, complete_cb: impl Into<Option<Job>>) -> &mut Self {
+        self.complete = Some(complete_cb.into());
+        self
+    }
+
+    /// Create a `Batch` with only `success` callback specified.
+    pub fn with_success_callback(&mut self, success_cb: Job) -> Batch {
```
jonhoo, line 166: I'd probably make this and the other methods take `self` so that they don't have to clone all the fields. Callers can always choose to `.clone()` before calling the method if they want to reuse the same builder values.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+
+    fn success(&mut self, success_cb: impl Into<Option<Job>>) -> &mut Self {
+        self.success = Some(success_cb.into());
+        self
+    }
+
+    fn complete(&mut self, complete_cb: impl Into<Option<Job>>) -> &mut Self {
+        self.complete = Some(complete_cb.into());
+        self
+    }
```
jonhoo, line 163: Why have these helper methods when they're one-liners and only used internally anyway?

jonhoo: I think you can get rid of them and then also skip them for the `#[builder]`.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+}
+
+pub struct BatchHandle<'a, S: Read + Write> {
+    bid: String,
+    prod: &'a mut Producer<S>,
+}
+
+impl<'a, S: Read + Write> BatchHandle<'a, S> {
+    /// ID issued by the Faktory server to this batch.
+    pub fn id(&self) -> &str {
```
jonhoo, line 194: We should really (in the next major version) introduce dedicated type for `JobId` and `BatchId` so that folks won't accidentally mix them up as they're both strings.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+        self.bid.as_ref()
+    }
+
+    pub(crate) fn new(bid: String, prod: &mut Producer<S>) -> BatchHandle<'_, S> {
+        BatchHandle { bid, prod }
+    }
+
+    /// Add the given job to the batch.
+    pub fn add(&mut self, mut job: Job) -> Result<(), Error> {
+        job.custom.insert("bid".into(), self.bid.clone().into());
```
jonhoo, line 204: Should we error if `job.custom` already contains `bid` for whatever reason?

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub fn start_batch(&mut self, mut batch: Batch) -> Result<BatchHandle<'_, S>, Error> {
+        batch.parent_bid = Some(self.bid.clone());
+        self.prod.start_batch(batch)
+    }
+
+    /// Commit this batch.
+    ///
+    /// The Faktory server will not queue any callbacks, unless the batch is committed.
+    /// Committing an empty batch will make the server queue the callback(s) right away.
+    /// Once committed, the batch can still be re-opened with [open_batch](struct.Producer.html#method.open_batch),
```
jonhoo, line 218:
```suggestion
    /// Once committed, the batch can still be re-opened with [open_batch](Producer::open_batch),
```
I believe this syntax (i.e., just naming the method with Rust naming) will also work.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+}
+
+/// Batch status retrieved from Faktory server.
+#[derive(Deserialize, Debug)]
+pub struct BatchStatus {
+    // Fields "bid", "created_at", "description", "total", "pending", and "failed"
+    // are described in the docs: https://github.com/contribsys/faktory/wiki/Ent-Batches#status
+    /// Id of this batch.
+    pub bid: String,
```
jonhoo, line 231: Might be cool to have a `BatchStatus::open` method that opens this `bid` as a handy shortcut?

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub failed: usize,
+
+    // The official golang client also mentions "parent_bid', "complete_st", and "success_st":
+    // https://github.com/contribsys/faktory/blob/main/client/batch.go#L8-L22
+    /// Id of the parent batch, provided this batch is a child ("nested") batch.
+    pub parent_bid: Option<String>,
+
+    /// State of the `complete` callback.
+    ///
+    /// See [complete](struct.Batch.html#structfield.complete).
```
jonhoo, line 255:
```suggestion
    /// See [complete](Batch::complete).
```
again, I _think_ this syntax should work.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,338 @@
...
+    #[serde(rename = "complete_st")]
+    pub complete_callback_state: String,
```
jonhoo, line 257: Can we provide a more strongly typed value here than `String`?

src/proto/mod.rs
```diff
@@ -175,6 +211,18 @@ impl<S: Read + Write> Client<S> {
             hello.labels = self.opts.labels.clone();
         }
 
+        single::write_command_and_await_ok(&mut self.stream, &hello)
+    }
+
+    #[cfg(feature = "ent")]
+    fn init_tracker(&mut self) -> Result<(), Error> {
```
jonhoo, line 218: Why have this as a separate function when it is (as far as I can tell?) exactly identical to `init`?

src/proto/single/ent.rs
```diff
@@ -1,6 +1,23 @@
...
+use serde::{
+    de::{Deserializer, IntoDeserializer},
+    Deserialize,
+};
 
-use crate::JobBuilder;
+use crate::{Error, JobBuilder};
+
+// Used to parse responses from Faktory that look like this:
+// '{"jid":"f7APFzrS2RZi9eaA","state":"unknown","updated_at":""}'
```
jonhoo, line 11: So am I right in assuming that the reason you need this helper is really just _specifically_ to handle the case where a datetime field is _set_  but _empty_? If so, let's be more explicit here. It wasn't clear to me reading this that you hadn't pasted the wrong thing and intended to paste a timestamp example :sweat_smile:

src/proto/single/ent.rs
```diff
@@ -74,6 +91,135 @@ impl JobBuilder {
     }
 }
 
+/// Info on job execution progress (retrieved).
+///
+/// The tracker is guaranteed to get the following details: the job's id (though
+/// they should know it beforehand in order to be ably to track the job), its last
+/// know state (e.g."enqueued", "working", "success", "unknown") and the date and time
```
jonhoo, line 98:
```suggestion
/// know state (e.g., "enqueued", "working", "success", "unknown"), and the date and time
```

src/proto/single/ent.rs
```diff
@@ -74,6 +91,135 @@ impl JobBuilder {
...
+/// ([desc](struct.ProgressUpdate.html#structfield.desc)) and completion percentage
+/// ([percent](struct.ProgressUpdate.html#structfield.percent)) may be available,
+/// if the worker provided those details.
+#[derive(Debug, Clone, Deserialize)]
+pub struct Progress {
+    /// Id of the tracked job.
+    pub jid: String,
+
+    /// Job's state.
+    pub state: String,
```
jonhoo, line 109: Can we provide a strongly-typed value for this field?

src/proto/single/ent.rs
```diff
@@ -74,6 +91,135 @@ impl JobBuilder {
...
+
+    /// When this job was last updated.
+    #[serde(deserialize_with = "parse_datetime")]
+    pub updated_at: Option<DateTime<Utc>>,
+
+    /// Percentage of the job's completion.
+    pub percent: Option<u8>,
```
jonhoo, line 116: Is this actually an integer and not a floating point number?

src/proto/single/ent.rs
```diff
@@ -74,6 +91,135 @@ impl JobBuilder {
...
+
+    /// Arbitrary description that may be useful to whoever is tracking the job's progress.
+    pub desc: Option<String>,
+}
+
+/// Info on job execution progress (sent).
+///
+/// In Enterprise Faktory, a client executing a job can report on the execution
+/// progress, provided the job is trackable. A trackable job is the one with "track":1
```
jonhoo, line 125:
```suggestion
/// progress, provided the job is trackable. A trackable job is one with "track":1
```

src/proto/single/ent.rs
```diff
@@ -74,6 +91,135 @@ impl JobBuilder {
...
+    pub percent: Option<u8>,
+
+    /// Arbitrary description that may be useful to whoever is tracking the job's progress.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[builder(default = "None")]
+    pub desc: Option<String>,
+
+    /// Allows to extend the job's reservation, if more time needed to execute it.
+    ///
+    /// Note that you cannot decrease the initial [reservation](struct.Job.html#structfield.reserve_for).
```
jonhoo, line 150: I'm not sure I follow this line?

src/proto/single/ent.rs
```diff
@@ -74,6 +91,135 @@ impl JobBuilder {
...
+
+impl ProgressUpdate {
+    /// Create a new instance of `ProgressUpdateBuilder` with job ID already set.
+    ///
+    /// Equivalent to creating a [new](struct.ProgressUpdateBuilder.html#method.new)
+    /// `ProgressUpdateBuilder`.
+    pub fn builder(jid: impl Into<String>) -> ProgressUpdateBuilder {
+        ProgressUpdateBuilder {
+            jid: Some(jid.into()),
+            ..ProgressUpdateBuilder::create_empty()
```
jonhoo, line 164: I assume `create_empty` isn't `pub`?

src/proto/single/ent.rs
```diff
@@ -74,6 +91,135 @@ impl JobBuilder {
...
+        }
+    }
+
+    /// Create a new instance of `ProgressUpdate`.
+    ///
+    /// While job ID is specified at `ProgressUpdate`'s creation time,
+    /// the rest of the [fields](struct.ProgressUpdate.html) are defaulted to _None_.
+    pub fn new(jid: impl Into<String>) -> ProgressUpdate {
+        ProgressUpdateBuilder::new(jid).build()
+    }
```
jonhoo, line 174: Why is this helper useful? I feel like we should either make people always use the builder _or_ have the fields be `pub` and settable after the fact, not both.

May be useful to have a shortcut like this though:
```rust
pub fn set_progress(jid: impl Into<String>), percent: u8) -> ProgressUpdate
```

src/proto/single/utils.rs
```diff
@@ -1,33 +1,41 @@
...
         .collect()
 }
 
+pub fn gen_random_jid() -> String {
+    gen_random_id(JOB_ID_LENGTH)
+}
+
+pub fn gen_random_wid() -> String {
+    gen_random_id(WORKER_ID_LENGTH)
+}
```
jonhoo, line 20: These feel like they should only be `pub(crate)`.

src/proto/single/mod.rs
```diff
@@ -182,6 +183,24 @@ impl JobBuilder {
...
+    ///
+    /// let _job = JobBuilder::new("order")
+    ///     .args(vec!["ISBN-13:9781718501850"])
+    ///     .build_trackable();
+    /// ```
+    #[cfg(feature = "ent")]
+    pub fn build_trackable(&mut self) -> Job {
+        self.add_to_custom_data("track", 1);
+        self.build()
+    }
```
jonhoo, line 203: What do you think of instead just having
```rust
if cfg!(feature = "ent") {
    self.add_to_custom_data("track", 1);
}
```
inside the regular `build` function? Is there ever a reason to _not_ mark a job as trackable?

If not, I would probably not make this a separate build method, and instead just have it be a builder-style method like:
```rust
impl JobBuilder {
    pub fn trackable(self, yes: bool) -> Self
}
```

src/proto/single/resp.rs
```diff
@@ -59,6 +59,32 @@ pub fn read_json<R: BufRead, T: serde::de::DeserializeOwned>(r: R) -> Result<Opt
...
+            expected: "non-empty blob representation of batch id",
+            received: "empty blob".into(),
+        }
+        .into()),
+        RawResponse::Blob(ref b) => Ok(std::str::from_utf8(b)
+            .map_err(|_| error::Protocol::BadType {
+                expected: "valid blob representation of batch id",
+                received: "unprocessable blob".into(),
+            })?
+            .into()),
```
jonhoo, line 81: Will the server ever actually send a batch ID as a blob? If not, we should error on it.

src/tracker/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+/// ```no_run
+/// use faktory::Tracker;
+/// let bid = String::from("W8qyVle9vXzUWQOg");
+/// let mut tracker = Tracker::connect(None)?;
+/// if let Some(status) = tracker.get_batch_status(bid)? {
+///     println!("This batch created at {}", status.created_at);
+/// }
+/// # Ok::<(), faktory::Error>(())
+/// ```
+pub struct Tracker<S: Read + Write> {
```
jonhoo, line 47: Why have a separate type for this rather than just having these be methods on `Client`?

jonhoo (changes requested):
Awesome stuff, thank you! Left a few notes, but none that I think are particularly onerous :tada:

src/producer/mod.rs
```diff
@@ -1,5 +1,13 @@
 use crate::error::Error;
-use crate::proto::{self, Client, Info, Job, Push, QueueAction, QueueControl};
+use crate::proto::{
+    self, parse_provided_or_from_env, Client, Info, Job, Push, QueueAction, QueueControl,
+};
+
+#[cfg(feature = "ent")]
+use crate::proto::{BatchHandle, CommitBatch, OpenBatch};
+#[cfg(feature = "ent")]
+use crate::Batch;
```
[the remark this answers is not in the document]
rustworthy: If there is, it's a mysterious one. I should have combined them, just like in the `proto` imports in [here](https://github.com/jonhoo/faktory-rs/pull/48#discussion_r1468851278)

src/producer/mod.rs
rustworthy: If the e2e tests, we are expecting the provided `bid` to be equal to the one returned. But now I am not sure why would do so in the first place. I guess it was just out of common sense (or out of desire to assert something :)), BUT this is not something they guarantee in the [docs](https://github.com/contribsys/faktory/wiki/Ent-Batches#batch-open-bid). At the same time, in the example in the docs they are using the _provided_ bid (rather than a returned one), when adding extra jobs to the re-opened batch. 

In the go [client](https://github.com/contribsys/faktory/blob/main/client/batch.go#L187-L203),
they are using the returned one and are not running the `==` check.

So the options are:
1) Throw away the returned bid and use the provided one for the extra jobs;
2) Follow the go bindings approach;

But I will definitely need to remove those assertions from the e2e tests.

src/proto/single/resp.rs
rustworthy: Turned out it always sends it as a blob. It was just too tempting to include both options.

src/producer/mod.rs
rustworthy: I just realized I never tested this case. Checked it: trying to open a never-actually-created batch with id, say, `non-existent-batch-id` will return an error `Protocol(Internal { msg: "No such batch non-existent-batch-id" })`. Will  look into it..

src/proto/batch/mod.rs
rustworthy: Uppercased, but looks like `struct.Tracker.html` link will still be needed since `Tracker` is not in the scope.

src/proto/batch/mod.rs
rustworthy: Now using `Parent batch description` and `Child batch description` for batch descriptions, and also simply `job_type` and `another_job_type`. But yes, I wanted it to be 'real-life' examples, maybe not the best choices for the docs.

src/proto/batch/mod.rs
rustworthy: I like the latter option better, just to avoid making it opinionated

src/proto/batch/mod.rs
rustworthy: shameless `cp` from the tests, I guess :(  Updated

src/proto/batch/mod.rs
rustworthy: There is a [config option](https://docs.rs/derive_builder/latest/derive_builder/#owned-aka-consuming) in `batch_builder` precisely for that.

src/producer/mod.rs
rustworthy: Add `maybe_bid` method on the `ReadToken` to handle this case

src/proto/batch/mod.rs
rustworthy: Indeed, they can be simply be skipped on the `BatchBuilder` and assigned directly onto the `Batch` in those terminal methods.

src/proto/batch/mod.rs
rustworthy: Iterating on it again, I believe the `description` setter should be custom and _really_ should have an `impl Into<String>` so that on the surface it's:
```rust
let batch = Batch::builder()
    .description("Batch description")
    .with_complete_callback(job_cb);
```

src/proto/batch/mod.rs
rustworthy:
```rust
let b = BatchBuilder::new().description("Batch description");
let _batch_with_complete_cb = b.clone().with_complete_callback(Job::builder("jt").build());
let _batch_with_success_cb = b.with_success_callback(Job::builder("jt").build());
```

src/proto/batch/mod.rs
rustworthy: Looks like swithing to async is  perfect timing. Added to [TODOs](https://github.com/jonhoo/faktory-rs/pull/49#issue-2086696629)

src/proto/batch/mod.rs
rustworthy: I can imagine this situation, where they decided they need (or they think they are _expected_) to somehow attribute the job to this batch and so put a `bid` into the custom hash, but say a wrong one. My understanding - we should just silently overwrite it and continue.

src/proto/batch/mod.rs
rustworthy: I've been thinking it over, since it's so tempting to enumerate those states, but I never found any promises on this part. Will check once again

src/proto/mod.rs
rustworthy: There is a whole block in `init` dedicated to a non-producer client (to a consumer basically), where we populate postname, wid etc fields, which is not needed for the `tracker`, since a it's not a producer, neither a consumer.

src/proto/single/ent.rs
rustworthy: Correct, specifically for this case: empty string :)

src/proto/single/ent.rs
rustworthy: They use ints across the doc [examples](https://github.com/contribsys/faktory/wiki/Ent-Tracking#setup).

The go client has got [`int`](https://github.com/contribsys/faktory/blob/main/client/tracking.go#L41) for this field.

src/proto/single/ent.rs
rustworthy: From the docs:
```
reserve_until allows the job to extend its current reservation until the given RFC3339 timestamp. You cannot make it sooner than the initial reservation so there's no point in setting this unless you know the job is long-running and needs more time
```
I guess "shorten" is right word, I'll also be more verbose:
```
Note that you cannot shorten the initial [reservation](struct.Job.html#structfield.reserve_for) via specifying an instant that is sooner than the job's initial reservation deadline.
```

src/proto/single/ent.rs
rustworthy: It is not. But those are generated methods and worth double-checking.
image

src/proto/batch/mod.rs
rustworthy: Yes, but should not have been pointing to the private field in the first place. Update the doc with the `struct.BatchBuilder.html#method.with_complete_callback` link

src/proto/batch/mod.rs
rustworthy: It works just fine! I also forgot to make `BatchHandle` public.

src/proto/batch/mod.rs
rustworthy: Confirmation from the Faktory's maintainer will be needed here. Emailed to Mike Perham and CC'd you.

src/proto/batch/mod.rs
rustworthy: Added this method. So they can go with:
```
let mut t = Tracker::connect(Some(&url)).unwrap();
let status = t.get_batch_status(bid.clone()).unwrap().unwrap();
let mut b = status.open(&mut p).unwrap().unwrap(); // using the bid from the retrieved BatchStatus
b.add(extra_job).unwrap();
```

src/proto/single/ent.rs
rustworthy: Indeed, progress update without an actual update does not make any sense (apart from the fact that the worker _did_  send an update, like a heart-beat - for the server to know the job is not dead and refresh the 
'updated_at' property).

Added the suggested shortcut plus another one - on the Progress:
```rust
pub fn update(&self, percent: u8) -> ProgressUpdate {
    set_progress(&self.jid, percent)
}
```
To easily create an update if they see that the percentage last known by the server for this job is out of date, e.g.:
```rust
if progress.percent != Some(100) {
    let upd = progress.update(100);
    assert!(t.lock().unwrap().set_progress(upd).is_ok())
}
```

rustworthy: ... and removed the redundant `new`

rustworthy: EDIT:

After playing with it as an end-user in the e2e tests, I realized that I'd like to have a method `update_percent` which will only update the job completion percentage and will copy over the description, whereas I want to update everything, I could use `update_builder` method returning an instance of `ProgressUpdateBuilder`:
```rust
    pub fn update_percent(&self, percent: u8) -> ProgressUpdate {
        ProgressUpdate::builder(&self.jid)
            .desc(self.desc.clone())
            .percent(percent)
            .build()
    }

    pub fn update_builder(&self) -> ProgressUpdateBuilder {
        ProgressUpdateBuilder::new(&self.jid)
    }
```

src/proto/single/mod.rs
rustworthy: Not that I can come up with, but people _may_ want to have some of the jobs untrackable.. As for the terminal method, I just wanted to save them a method call: instead of `another_method().trackable().build()` simply `another_method().build_trackable()`.

After thinking it over a little bit, I realized that those who have paid for the Ent Faktory subscription will want all the paid features included by default and should not opt in additionally. I got rid of the `build_trackable` and  updated the logic and tests accordingly. Also updated the docs for `build` mentioning how they can opt out of tracking.

src/tracker/mod.rs
rustworthy: I did not like a `Producer` to be able to set job progress, and I also don't think that the Producer is the only one who can be interested in the job/bath execution progress. Image we have a process running exclusively for tracking purposes: instantiating a `Producer` (not to mention a `Consumer`) for these purposes just feels weird. Hence separation.

I also believe that the `Client` should be responsible for connecting, reconnecting and issuing commands, and should not know anything about setting and getting progress or batch statuses.

That was my reasoning.

rustworthy:
> Awesome stuff, thank you! Left a few notes, but none that I think are particularly onerous 🎉

I hope all the threads have been addressed. One thing is [pending](https://github.com/jonhoo/faktory-rs/pull/48#discussion_r1468856234) though, but as soon as confirmation from Mike is there, I will cover it with great pleasure.

src/consumer/mod.rs
```diff
@@ -1,5 +1,9 @@
 use crate::error::Error;
-use crate::proto::{self, Client, ClientOptions, HeartbeatStatus, Reconnect};
+
+use crate::proto::{
```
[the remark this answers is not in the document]
jonhoo: Heh, yeah, there's a lot of debate around exactly how these should be grouped. My current take is that without perfect automatic grouping separation, the best way forward is to just group everything together and sort alphabetically. After all, the import list isn't something people usually end up looking at manually.

I don't feel too strongly about trying to enforce this in CI since, as you say, the options are currently unstable. But would be okay with adding this as an additional step to the existing `cargo fmt` job if you feel strongly about it :+1:

src/producer/mod.rs
jonhoo: I'm generally inclined to follow the go bindings where there is ambiguity, since that is (as far as I can tell) the reference implementation maintained by the Faktory folks.

src/proto/batch/mod.rs
jonhoo: I like real-life examples, but I think they then belong in an actual example (say, in `examples/`) with more context + end-to-end code. For doc code for individual methods (without that context), it just reads kind of strange.

src/proto/batch/mod.rs
jonhoo: Ah, a nice trick here is:
```rust
#[cfg(doc)]
use whatever::Tracker;
```
to bring it in scope so rustdoc can find it.

src/proto/batch/mod.rs
jonhoo: How about making the `Ok` type be `Option<Bid>`, so that we can return the old `Bid` if there was one? The user can always choose to ignore the old value if they don't care.

src/proto/mod.rs
jonhoo: Ohh, I see. That's fine. I just figured it might be an opportunity to share some code, but not terribly important in this case.

src/proto/single/ent.rs
jonhoo: Isn't the first of those just `self.clone().percent(percent)`?

src/producer/mod.rs
```diff
@@ -1,5 +1,9 @@
 use crate::error::Error;
-use crate::proto::{self, Client, Info, Job, Push, QueueAction, QueueControl};
+use crate::proto::{
+    self, parse_provided_or_from_env, Client, Info, Job, Push, QueueAction, QueueControl,
+};
+#[cfg(feature = "ent")]
+use crate::proto::{Batch, BatchHandle, CommitBatch, OpenBatch};
```
jonhoo, line 6: The `cfg`-gated `use` should probably be in their own section, otherwise the sorting gets a little weird.

src/producer/mod.rs
```diff
@@ -129,6 +130,33 @@ impl<S: Read + Write> Producer<S> {
...
+    /// This will not error if a batch with the provided `bid` does not exist,
+    /// rather `Ok(None)` will be returned.
+    #[cfg(feature = "ent")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "ent")))]
+    pub fn open_batch(&mut self, bid: String) -> Result<Option<BatchHandle<'_, S>>, Error> {
+        let bid = self.c.issue(&OpenBatch::from(bid))?.maybe_bid()?;
+        match bid {
+            Some(bid) => Ok(Some(BatchHandle::new(bid, self))),
+            None => Ok(None),
+        }
```
jonhoo, line 153:
```suggestion
        Ok(bid.map(|bid| BatchHandle::new(bid, self)))
```

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+
+impl Batch {
+    /// Create a new `BatchBuilder`.
+    pub fn builder() -> BatchBuilder {
+        BatchBuilder::new()
+    }
+}
+
+impl BatchBuilder {
+    fn build(self) -> Batch {
```
jonhoo, line 150: As currently written, it is not possible to construct a batch that has neither a success or a complete callback. Is that intentional?

src/proto/single/ent.rs
```diff
@@ -67,13 +84,198 @@ impl JobBuilder {
...
+    pub fn new(jid: impl Into<String>) -> ProgressUpdateBuilder {
+        ProgressUpdateBuilder {
+            jid: Some(jid.into()),
+            ..ProgressUpdateBuilder::create_empty()
+        }
+    }
+}
+
+/// Create an instance of `ProgressUpdate` for the job specifying its completion percentage.
+pub fn set_progress(jid: impl Into<String>, percent: u8) -> ProgressUpdate {
```
jonhoo, line 242: How about making this a method on `ProgressUpdate` that doesn't take `self`, so that it can be called as `ProgressUpdate::set(jid, percent)`?

src/proto/single/ent.rs
```diff
@@ -67,13 +84,198 @@ impl JobBuilder {
...
+    pub fn builder(jid: impl Into<String>) -> ProgressUpdateBuilder {
+        ProgressUpdateBuilder::new(jid)
+    }
+}
+
+impl ProgressUpdateBuilder {
+    /// Builds an instance of ProgressUpdate.
+    pub fn build(&self) -> ProgressUpdate {
+        self.try_build()
+            .expect("All required fields have been set.")
```
jonhoo, line 229:
```suggestion
            .expect("Only jid is required, and it is set by all constructors.")
```

src/proto/single/mod.rs
```diff
@@ -178,7 +180,16 @@ impl JobBuilder {
     }
 
     /// Builds a new [`Job`] from the parameters of this builder.
-    pub fn build(&self) -> Job {
+    ///
+    /// For Enterprise edition of Faktory builds a new _trackable_ `Job`.
+    /// In Enterprise Faktory, a progress update can be sent and received only for the jobs
+    /// that have been explicitly marked as trackable via `"track":1` in the job's custom hash.
+    /// In case you have a reason to opt out of tracking, either unset (remove) the "track" on
+    /// the resulted job's [`custom`](Job::custom) hash or set to 0.
```
jonhoo, line 188:
```suggestion
    /// the resulted job's [`custom`](Job::custom) hash or set it to 0.
```

src/tracker/mod.rs
jonhoo: This makes me wonder if we should make `Client` public, and have these methods be directly on `Client`. What do you think?

src/proto/single/ent.rs
```diff
@@ -67,13 +84,198 @@ impl JobBuilder {
...
+
+/// Info on job execution progress (retrieved).
+///
+/// The tracker is guaranteed to get the following details: the job's id (though they should
+/// know it beforehand in order to be ably to track the job), its last known [`state`](JobState), and
+/// the date and time the job was last updated. Additionally, arbitrary information on what's going
+/// on with the job ([`desc`](ProgressUpdate::desc)) and the job's completion percentage
+/// ([`percent`](ProgressUpdate::percent)) may be available, if the worker has provided those details.
+#[derive(Debug, Clone, Deserialize)]
+pub struct Progress {
```
jonhoo, line 145: This file is getting pretty long. I wonder if we should start splitting things into submodules for individual features (like `mod progress`). Please do that in a separate PR though so we avoid holding this one up longer :+1:

jonhoo (changes requested):
We're getting pretty close! Just a few more notes, most of which are minor. I've also gone through and marked all the addressed threads as resolved :+1: 

One request: please avoid force-pushing until at the very end when all is approved, otherwise it makes reviewing the diffs a fair bit harder (mostly due to shortcomings in GitHub) :sweat_smile:

src/consumer/mod.rs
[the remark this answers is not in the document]
rustworthy: I believe a helper command plus an extra CI check would be nice to have, but definitely not in this PR: I ran `cargo +nightly fmt -- --config group_imports=one` on the sources and some of the modules got formatter - this would be a distracting diff.

src/proto/batch/mod.rs
rustworthy: This is clever. Updated

src/proto/batch/mod.rs
rustworthy: Which is effectively sticking to the [`HashMap`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert) API. I see your point!

rustworthy: Also added a check to `test_batch_of_jobs_can_be_initiated` test:
```rust
assert!(b.add(job_1).unwrap().is_none());
assert!(b.add(job_2).unwrap().is_none());
assert_eq!(b.add(job_3).unwrap().unwrap(), serde_json::Value("check-check"));
```
where "check-check" is the value of the bid added to `job_3` prior to adding this job to the batch.

Of notice here, is that we are returning an option of serde_json::Value since this is how we keep data in the Job's custom hash. Interestingly, the user is free to add a key `bid` and whatever (implementing into Value of course) value via `add_to_custom_hash`. If they do so, we will return them back their value.

Thinking it over, I can imagine a situation their some domain logic will have a  `bid` which does not stand for batch ID, rather something else - whatever. This is a collision. So, returning a Value that was stored in the hash under the `bid` key will _at least_ attract some attention to this. Also adding to the method's doc of course.

src/proto/batch/mod.rs
rustworthy: After receiving confirmation from the Faktory maintainers that they are not planning to change the API, enumerated those states via `CallbackState`, so we've got:
```rust
let status = tracker.get_batch_status(bid)?.unwrap();
match status.complete_callback_state {
    CallbackState::Pending => {},
    CallbackState::Enqueued => {},
    CallbackState::FinishedOk => {},
}
```
One can wonder, why there is no `FinshedNotOk` variant. I've updated one of the tests (`test_callback_will_be_queued_upon_commit_even_if_batch_is_empty`) to demonstrate that if the callback has been executed with a failure, it's state just stays `CallbackState::Enqueued`.

src/tracker/mod.rs
rustworthy: I've worked with the Go client the other day when porting one of our integration tests. I remember I got stuck for just a little while where I needed to consume the jobs I had previously published. And yes it turned out I could just use the same construct I had used for pushing those jobs.

I see, the idea is pretty straightforward: you instantiate a client and perform _all_ operations with its help. But I guess this just a different approach, and I am not sure I am a fan of it. I also understand I can be biased :smile: 

I've also just checked the API of the faktory python [bindings](https://github.com/cdrx/faktory_worker_python?tab=readme-ov-file#worker-example). They've  exposing:  `client` (our Producer), `server`, and `Worker` (our Consumer). They are implementing it exactly according to the Faktory pieces description that one can find [here](https://github.com/contribsys/faktory_worker_ruby?tab=readme-ov-file#faktory_worker_ruby) for example. They are using a similar approach(with `faktory.work()`) in [node](https://github.com/jbielick/faktory_worker_node?tab=readme-ov-file#processing-jobs) and [C#](https://github.com/tpod/FaktoryWorker?tab=readme-ov-file#worker) bindings. But for me the `Client` vs `Worker` is a little bit confusing. Also notice how this is _different_ from the Go client API, where there is simply `Client`.

I've worked previously with [Kafka.js](https://kafka.js.org/docs/getting-started) lib where the concept of `producer` and `consumer`  (which is kind of projected from `Kafka` to the bindings) just felt so natural to me.  In most of the cases, you are launching a process to either produce, or consume, so you only need role-specific methods.

src/proto/batch/mod.rs
rustworthy: It is. The batch should have either/both callbacks as per [docs](https://github.com/contribsys/faktory/wiki/Ent-Batches#batch-new-json). 

Interestingly, the Go client will not force the user to specify either/both of them, committing such a batch will make the server error back with:
```
ERR Batch must define a success or complete callback
```

I learned it accidentally: I just forgot to do so initially when writing this [test](https://github.com/rustworthy/faktory/actions/runs/7765783083/job/21180846057)

src/proto/single/ent.rs
rustworthy: Yeah I like it better when it's in the `ProgressUpdate` namespace!

src/proto/single/ent.rs
rustworthy: Sure! The only thing is we are done with Ent Faktory features :smile:

The rest of the enterprise features:
- [Ent Cron](https://github.com/contribsys/faktory/wiki/Ent-Cron)
- [Ent Metrics](https://github.com/contribsys/faktory/wiki/Ent-Metrics) 
- [Ent Redis Gateway](https://github.com/contribsys/faktory/wiki/Ent-Redis-Gateway)
- [Ent Throttling](https://github.com/contribsys/faktory/wiki/Ent-Throttling)
- [Ent Remote Redis](https://github.com/contribsys/faktory/wiki/Ent-Remote-Redis)

are all deployment related, with most of them fine-tuned via config files in `/etc/faktory/conf.d` on the machine.

So I suggest we split it when they introduce new features?

rustworthy: ... although I hope new features will be for the community edition :smile:

tests/real/enterprise.rs
```diff
@@ -370,3 +391,693 @@ fn ent_unique_job_bypass_unique_lock() {
...
+    match s.complete_callback_state {
+        CallbackState::Enqueued => {}
+        _ => panic!("Expected the callback to have been enqueued"),
+    }
+    assert_eq!(s.pending, 2); // ... though there are pending jobs
+    assert_eq!(s.total, 2);
+
+    // Subtest 4 result:
+    // We were able to add more jobs to the batch for which the Faktory server had already
+    // queued the callback.
```
rustworthy, line 1071: The other option would be to not make any assertions here, but since this is a documented guarantee, I decided to leave this subtest as-is and as soon as the fixes are introduced - to update the test.

rustworthy, line 1071: The other option would be to not make any assertions here, but since this is a documented guarantee, I decided to leave this subtest as-is and as soon as the fixes are introduced - to update the test.

src/tracker/mod.rs
jonhoo: Ah, so, to be clear, what I'm proposing is that we have a `Client` type, but that you _cannot call any methods on it_ (except maybe the tracker methods). Then we have a `Producer` and `Consumer` type, both of which wrap a `Client` and adds additional methods.

src/proto/single/ent.rs
jonhoo: I think we can probably split even now given how long the file is. But a weaker preference for sure if we're already basically done :p

src/proto/batch/mod.rs
```diff
@@ -235,6 +239,34 @@ impl<'a, S: Read + Write> BatchHandle<'a, S> {
     }
 }
 
+// Not documented, but existing de fakto and also mentioned in the official client
+// https://github.com/contribsys/faktory/blob/main/client/batch.go#L17-L19
+/// State of a `callback` job of a [`Batch`].
+#[derive(Debug, Clone, Deserialize)]
+pub enum CallbackState {
```
jonhoo, line 253: If we suspect this may grow to include something like a `FinishedNotOk` in the future, we could mark this with `#[non_exhaustive]` so that we _can_ add variants to it. The downside is that callers don't get exhaustive matching, which is also pretty sad.

src/proto/batch/mod.rs
```diff
@@ -235,6 +239,34 @@ impl<'a, S: Read + Write> BatchHandle<'a, S> {
     }
 }
 
+// Not documented, but existing de fakto and also mentioned in the official client
+// https://github.com/contribsys/faktory/blob/main/client/batch.go#L17-L19
+/// State of a `callback` job of a [`Batch`].
+#[derive(Debug, Clone, Deserialize)]
```
jonhoo, line 252: Let's add `Copy`, `PartialEq`, and `Eq`.

tests/real/enterprise.rs
```diff
@@ -608,8 +608,8 @@ fn test_batch_of_jobs_can_be_initiated() {
     assert_eq!(s.failed, 0);
     // Docs do not mention it, but the golang client does:
     // https://github.com/contribsys/faktory/blob/main/client/batch.go#L17-L19
-    assert_eq!(s.success_callback_state, ""); // we did not even provide the 'success' callback
-    assert_eq!(s.complete_callback_state, ""); // the 'complete' callback is pending
+    assert_eq!(s.success_callback_state.to_string(), "Pending"); // we did not even provide the 'success' callback
```
jonhoo, line 612: When you add `PartialEq` to this, you'll be able to just `assert_eq!` with `CallbackState::Pending` directly.

tests/real/enterprise.rs
```diff
@@ -919,7 +919,10 @@ fn test_callback_will_be_queued_upon_commit_even_if_batch_is_empty() {
     let s = t.get_batch_status(bid.clone()).unwrap().unwrap();
     assert_eq!(s.total, 0);
     assert_eq!(s.complete_callback_state.to_string(), "FinishedOk");
-    assert_eq!(s.success_callback_state.to_string(), "FinishedOk");
+    // Still `Enqueued` due to the fact that it was not finished with success.
+    // Had we registered a handler for `success_cb_jobtype` returing Ok(()) rather then Err(),
+    // the state would be `FinishedOk` just like it's the case with the `complete` callback.
```
jonhoo, line 926: I think this is worth adding to the docs on `CallbackState`.

tests/real/enterprise.rs
```diff
@@ -588,7 +588,7 @@ fn test_batch_of_jobs_can_be_initiated() {
 
     assert!(b.add(job_1).unwrap().is_none());
     assert!(b.add(job_2).unwrap().is_none());
-    assert_eq!(b.add(job_3).unwrap().unwrap(), "check-check");
+    assert_eq!(b.add(job_3).unwrap().unwrap(), Value::from("check-check"));
```
jonhoo, line 591: I _believe_ `serde_json::Value` implements `PartialEq<str>`, so the `Value::from` here _may_ not be needed.

jonhoo (changes requested):
Nearly there!

src/proto/batch/mod.rs
rustworthy: `FinishedNotOk` looks like a very likely candidate. Added `#[non_exhaustive]`

src/proto/batch/mod.rs
rustworthy: Also added them for `JobState`

tests/real/enterprise.rs
rustworthy: Done

tests/real/enterprise.rs
rustworthy: Indeed!

src/tracker/mod.rs
rustworthy: I see. Removed the tracker wrapper, exposed `connect` and `connect_with` on the `Client` directly + those `get_batch_status`, `get_progress` , and `set_progress`. 

I also refactored `is_producer` of the `ClientOptions` to be `is_worker` which solved my previous problem, where I did not want to use `new_producer` on the `Client` for the `Tracker`. 

We only want some special configuration for a consumer (worker in Faktory lingo). So `is_worker` will be `false` by default, and only have `true` when this client is being instantiated for `Consumer` purposes.

src/proto/batch/mod.rs
```diff
@@ -0,0 +1,428 @@
...
+    /// Once committed, the batch can still be re-opened with [open_batch](Producer::open_batch),
+    /// and extra jobs can be added to it.
+    pub fn commit(self) -> Result<(), Error> {
+        self.prod.commit_batch(self.bid)
+    }
+}
+
+// Not documented, but existing de fakto and also mentioned in the official client
+// https://github.com/contribsys/faktory/blob/main/client/batch.go#L17-L19
+/// State of a `callback` job of a [`Batch`].
```
jonhoo, line 251:
```suggestion
/// State of a `callback` job of a [`Batch`].
///
// Not documented, but existing de fakto and also mentioned in the official client
// https://github.com/contribsys/faktory/blob/main/client/batch.go#L17-L19
```
rustdoc expects the first paragraph of a doc comment to be the main "tagline" for the item being documented, and will often show _just_ that.

src/lib.rs
```diff
@@ -76,5 +76,11 @@ pub use tls::TlsStream;
...
-pub use crate::proto::Reconnect;
-pub use crate::proto::{Job, JobBuilder};
+pub use crate::proto::{Client, Job, JobBuilder, Reconnect};
+
+#[cfg(feature = "ent")]
+#[cfg_attr(docsrs, doc(cfg(feature = "ent")))]
+pub use crate::proto::{
+    Batch, BatchBuilder, BatchHandle, BatchStatus, CallbackState, JobState, Progress,
+    ProgressUpdate, ProgressUpdateBuilder,
+};
```
jonhoo, line 86: The root of the crate is starting to have a _lot_ of types in it. How would you feel about making this:
```suggestion
/// Constructs only available with the enterprise version of Faktory. 
pub mod ent {
    pub use crate::proto::{
        Batch, BatchBuilder, BatchHandle, BatchStatus, CallbackState, JobState, Progress,
        ProgressUpdate, ProgressUpdateBuilder,
    };
}
```

jonhoo: Or, alternatively, not exposing most of these at the root at all?

src/proto/mod.rs
```diff
@@ -72,10 +101,12 @@ pub(crate) struct ClientOptions {
     pub(crate) labels: Vec<String>,
 
     /// Password to authenticate with
-    /// Defaults to None,
+    /// Defaults to None.
     pub(crate) password: Option<String>,
 
-    is_producer: bool,
+    /// Whether this client is instatianted for
+    /// a consumer (`worker` in Faktory terms).
```
jonhoo, line 108:
```suggestion
    /// a consumer ("worker" in Faktory terms).
```

src/proto/mod.rs
```diff
@@ -86,12 +117,59 @@ impl Default for ClientOptions {
...
             labels: vec!["rust".to_string()],
             password: None,
-            is_producer: false,
+            is_worker: false,
         }
     }
 }
 
-pub(crate) struct Client<S: Read + Write> {
+/// A construct used internally by [`Producer`] and [`Consumer`].
```
jonhoo, line 125:
```suggestion
/// A Faktory connection that represents neither a [`Producer`] nor a [`Consumer`].
///
```

src/proto/mod.rs
```diff
@@ -86,12 +117,59 @@ impl Default for ClientOptions {
...
+/// Can be used directly for retrieving and updating information on a job's execution progress
+/// (see [`Progress`] and [`ProgressUpdate`]), as well for retrieving a batch's status
+/// from the Faktory server (see [`BatchStatus`]).
```
jonhoo, line 128:
```suggestion
/// Useful for retrieving and updating information on a job's execution progress
/// (see [`Progress`] and [`ProgressUpdate`]), as well for retrieving a batch's status
/// from the Faktory server (see [`BatchStatus`]).
```

src/proto/mod.rs
```diff
@@ -86,12 +117,59 @@ impl Default for ClientOptions {
...
+///         JobState::Success => {
+///         # /*
+///             ...
+///         # */
+///         },
+///         # /*
+///         ...
+///         # */
+///         # _ => {},
+///     }
```
jonhoo, line 146:
```suggestion
///     if let JobState::Success = progress.state {
///         # /*
///         ...
///         # */
///     }
```

src/proto/mod.rs
```diff
@@ -86,12 +117,59 @@ impl Default for ClientOptions {
...
+/// }
+/// # Ok::<(), faktory::Error>(())
+/// ```
+/// Sending an update on a job's execution progress:
```
jonhoo, line 150:
```suggestion
///
/// Sending an update on a job's execution progress:
///
```

src/proto/mod.rs
```diff
@@ -86,12 +117,59 @@ impl Default for ClientOptions {
...
+/// let jid = String::from("W8qyVle9vXzUWQOf");
+/// let mut cl = Client::connect(None)?;
+/// let progress = ProgressUpdateBuilder::new(&jid)
+///     .desc("Almost done...".to_owned())
+///     .percent(99)
+///     .build();
+/// cl.set_progress(progress)?;
+/// # Ok::<(), faktory::Error>(())
+///````
+/// Fetching a batch's status:
```
jonhoo, line 162:
```suggestion
///
/// Fetching a batch's status:
///
```

src/proto/mod.rs
```diff
@@ -193,6 +260,47 @@ impl<S: Read + Write> Drop for Client<S> {
...
+    }
+
+    /// Fetch information on a batch of jobs execution progress.
+    pub fn get_batch_status(&mut self, bid: String) -> Result<Option<BatchStatus>, Error> {
+        let cmd = GetBatchStatus::from(bid);
+        self.issue(&cmd)?.read_json()
+    }
+}
+
+impl Client<TcpStream> {
```
jonhoo, line 285: I recommend moving this as the first `impl` block on `Client` so that it is more clearly visible in the generated docs.

jonhoo (changes requested):
Basically just nits left!

One thought I have: is there even a reason to have `Producer` and `Client` be different types? Like, could we literally change `Producer` to
```rust
pub type Producer<S> = Client<S>;
```
?

src/proto/batch/mod.rs
rustworthy: Those two lines were not meant to be published.
image

With the suggested changes they will still be simple comments, but it kind of looks confusing?

src/lib.rs
rustworthy: I really like this namespacing!

src/proto/mod.rs
rustworthy: Yeah, I went crazy here :)

src/proto/batch/mod.rs
jonhoo: Ohh, I missed that they were normalt comments! Fine to leave them as is then 👍

src/lib.rs
rustworthy: Indeed, much better, less noise
image

src/proto/mod.rs
rustworthy: done

rustworthy:
> Basically just nits left!
> 
> One thought I have: is there even a reason to have `Producer` and `Client` be different types? Like, could we literally change `Producer` to
> 
> ```rust
> pub type Producer<S> = Client<S>;
> ```
> 
> ?
Hmm we could not without exposing methods `enqueue` , `info`,  `queue_pause`, `queue_resume`, `start_batch`, `open_batch` and `commit_batch`  on the `Client`.

jonhoo (approved):
Going to land this as that discussion is a bit separate, and would probably be breaking.

jonhoo:
Looks like https://github.com/jonhoo/faktory-rs/pull/54 caused some conflicts with this one — I've resolved them and pushed directly to your branch to merge :+1:

jonhoo:
Worth highlighting the change I made in https://github.com/jonhoo/faktory-rs/pull/48/commits/6c98127365fdf88c49410b92773c41214fdb46ee too. I also decided to do that because it meant my editor kept changing the `use`s, and then the job requires that I run a manual command to change them back, which became tedious pretty quickly :sweat_smile:

jonhoo:
Released in 0.12.5 :tada:

rustworthy:
> > Hmm we could not without exposing methods `enqueue` , `info`, `queue_pause`, `queue_resume`, `start_batch`, `open_batch` and `commit_batch` on the `Client`.
> 
> Right, but I guess my argument is that there isn't semantically really a difference between a `Client` and a `Producer` (I think?).

I guess semantically both Producer and Consumer are Clients talking to the Factory Server. The fact that Go bindings only use Client speaks for itself. At the same time - just a refresher - the Faktory lingo has got a Client Vs Worker. So, if we decide to merge client with producer, we may want to rename the Consumer to be Worker, wdyt?

I believe, though, the current setup IS semantically correct.

jonhoo:
Yeah, that's my thinking too: `Producer` -> `Client`, `Consumer` -> `Worker`, and have tracking just be additional methods on `Client`. After all, there's not _really_ any difference between a client that submits jobs and one that tracks them.

rustworthy:
> Worth highlighting the change I made in [6c98127](https://github.com/jonhoo/faktory-rs/commit/6c98127365fdf88c49410b92773c41214fdb46ee) too. I also decided to do that because it meant my editor kept changing the `use`s, and then the job requires that I run a manual command to change them back, which became tedious pretty quickly 😅

Hmm the idea was to never have to return to the question of 'use's grouping :), where the pipeline job would be the guard, rather than a reviewer. Anyhow, it was 1) a nit and 2) a nice nightly feature to learn about :microscope:

rustworthy:
> Yeah, that's my thinking too: `Producer` -> `Client`, `Consumer` -> `Worker`, and have tracking just be additional methods on `Client`. After all, there's not _really_ any difference between a client that submits jobs and one that tracks them.

Will be addressed in #49

PR #51, opened 2024-01-31
Refactor Consumer Module: Introduce JobRunner Trait for Flexible Job Handling

Kab1r:
## Background
I need to have access to some initialized configuration / state for my Handler.
While I could use `once_cell` to initialize static variables to hold a configuration, it isn't particularly idiomatic.

## Summary of Changes
This PR introduces a change to the `Consumer` module, converting the `JobRunner` type into a trait.
The `JobRunner` trait is now implemented for types that can handle jobs, providing more flexibility for handling jobs within a `Consumer`.

### Example Usage

```rust
use faktory::{ConsumerBuilder, JobRunner, Job};
use std::io;

struct MyHandler {
    config: String,
}
impl JobRunner<io::Error> for MyHandler {
   fn run(&self, job: Job) -> Result<(), io::Error> {
      println!("config: {}", self.config);
      println!("job: {:?}", job);
      Ok(())
  }
}

let mut c = ConsumerBuilder::default();
let handler = MyHandler {
   config: "foo".to_string(),
};
c.register_runner("bar", handler);
let mut c = c.connect(None).unwrap();
if let Err(e) = c.run(&["default"]) {
    println!("worker failed: {}", e);
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/51)


src/consumer/mod.rs
```diff
@@ -12,8 +12,53 @@ const STATUS_RUNNING: usize = 0;
...
+}
+// Implements JobRunner for a closure that takes a Job and returns a Result<(), E>
+impl<E, F> JobRunner<E> for F
+where
+    F: Fn(Job) -> Result<(), E> + Send + Sync,
+{
+    fn run(&self, job: Job) -> Result<(), E> {
+        self(job)
+    }
+}
```
Kab1r, line 60: The previous `JobRunner` Type:
```rust
dyn Fn(Job) -> Result<(), E> + Send + Sync
```
implements our new `JobRunner` Trait

src/consumer/mod.rs
```diff
@@ -12,8 +12,53 @@ const STATUS_RUNNING: usize = 0;
...
+/// if let Err(e) = c.run(&["default"]) {
+///     println!("worker failed: {}", e);
+/// }
+/// ```
+pub trait JobRunner<E>: Send + Sync {
+    /// A handler function that runs a job.
+    fn run(&self, job: Job) -> Result<(), E>;
+}
+// Implements JobRunner for a closure that takes a Job and returns a Result<(), E>
+impl<E, F> JobRunner<E> for F
```
jonhoo, line 53: Unfortunately I think this won't do quite what you want — because this is a blanket implementation, no other implementations will be allowed, so no-one will be able to `impl JobRunner` themselves. It sucks, but it's a reality of the way the orphan rules in Rust works. I think the way we'll have to do this instead is to implement the trait just for `Box<dyn JobRunner>` (i.e., not as a blanket impl), leave `register` the way it currently is, and then add a `register_runner` method that takes `R: JobRunner`.

src/consumer/mod.rs
```diff
@@ -12,8 +12,53 @@ const STATUS_RUNNING: usize = 0;
...
+/// let handler = MyHandler {
+///    config: "bar".to_string(),
+/// };
+/// c.register("foo", handler);
+/// let mut c = c.connect(None).unwrap();
+/// if let Err(e) = c.run(&["default"]) {
+///     println!("worker failed: {}", e);
+/// }
+/// ```
+pub trait JobRunner<E>: Send + Sync {
```
jonhoo, line 48: I think the `E` here would be better as an associated type than as a generic type parameter. After all, it's not likely that a given implementor will be able to support an arbitrary error type, and instead the implementing type will be specific for a given error type.

README.md
```diff
@@ -52,7 +52,7 @@ If you want to **accept** jobs from Faktory, use `Consumer`.
 use faktory::ConsumerBuilder;
 use std::io;
 let mut c = ConsumerBuilder::default();
-c.register("foobar", |job| -> io::Result<()> {
+c.register("foobar", |job: Job| -> io::Result<()> {
```
jonhoo, line 55: The fact that this annotation is now required everywhere is an indicator that this change as-is would be a breaking change. See the comments later though which I think will allow us to leave this as-is and avoid the breaking change.

jonhoo (changes requested):
I totally see the need! Left some comments on how to achieve it though.

src/consumer/mod.rs
Kab1r: Correct me if I am wrong, but my understanding of the orphan rules was that either the trait or the type for a given `impl` had to be in the crate.
You should be able to implement the trait for any type that is declared in the same crate as the implementation of the trait.
I would not expect that a user should be able to implement the `JobRunner` for a closure manually.

src/consumer/mod.rs
```diff
@@ -12,8 +12,56 @@ const STATUS_RUNNING: usize = 0;
...
+    type E;
+    /// A handler function that runs a job.
+    fn run(&self, job: Job) -> Result<(), Self::E>;
+}
+// Implements JobRunner for a closure that takes a Job and returns a Result<(), E>
+impl<E, F> JobRunner for F
+where
+    F: Fn(Job) -> Result<(), E> + Send + Sync,
+{
+    type E = E;
```
Kab1r, line 59: The `register` method any type `E`for closures, so I have not constrained the type to `std::error::Error`

src/consumer/mod.rs
jonhoo: This led me down quite a rabbit hole! It wasn't actually the orphan rules I was concerned about (mis-typed), it was the overlapping impls rules. Specifically, I was concerned that having a blanket impl here would prevent any downstream crate from implementing the trait since the impl would be potentially-conflicting with the blanket impl. But, it turns out, Rust doesn't consider this a hazard for `Fn` specifically because `Fn`'s set of implementations are "closed" — that is, the Rust standard library has essentially committed to not automatically implementing `Fn` for any new types. This is not the case for all traits — for instance, if you `s/Fn/Unpin/` (just as an example exercise), this blanket impl _does_ prevent implementing the trait for any other type, precisely because std reserves the right to `impl Unpin` automatically for user types in future versions. Interestingly, using a trait defined in the current crate as a bound is _also_ considered okay, because arriving at a conflict would either require that we add a blanket impl of that second trait (which is a breaking change), or the consuming crate would need to be the one to impl the second trait themselves, at which point they'll be told about the conflict.

All that to say, this is indeed okay, it just turns out to be for a surprisingly complex reason :sweat_smile:

src/consumer/mod.rs
```diff
@@ -199,6 +247,19 @@ impl<E> ConsumerBuilder<E> {
         self
     }
 
+    /// Register a handler for the given job type (`kind`).
+    ///
+    /// Whenever a job whose type matches `kind` is fetched from the Faktory, the given handler
+    /// object is called with that job as its argument.
+    pub fn register_runner<K, H>(&mut self, kind: K, runner: H) -> &mut Self
```
jonhoo, line 254: I think you can change `register` to just call this method internally. We only really need it to give the type checker a boost to realize the required closure type.

jonhoo (changes requested):
Almost there — just a couple of nits left!

src/consumer/mod.rs
```diff
@@ -12,8 +12,56 @@ const STATUS_RUNNING: usize = 0;
...
+/// let mut c = c.connect(None).unwrap();
+/// if let Err(e) = c.run(&["default"]) {
+///     println!("worker failed: {}", e);
+/// }
+/// ```
+pub trait JobRunner: Send + Sync {
+    /// The error type that the handler may return.
+    type E;
+    /// A handler function that runs a job.
+    fn run(&self, job: Job) -> Result<(), Self::E>;
```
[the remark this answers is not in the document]
Kab1r: I haven't quite figured out this yet, I think I need to understand the rules about overlapping blanket implementations more closely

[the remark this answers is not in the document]
Kab1r: I think I have this done, the `BoxRunner` trait is implemented for `Box<F>`, `&F`, and `&mut F`,
where `F: Fn(Job) -> Result<(), E> + Send + Sync`.
It is no longer implemented for `F` because it appears to overlap with all the other implementations.

src/consumer/mod.rs
```diff
@@ -243,7 +263,7 @@ impl<E> ConsumerBuilder<E> {
         K: Into<String>,
         H: Fn(Job) -> Result<(), E> + Send + Sync + 'static,
     {
-        self.register_runner(kind, handler)
+        self.register_runner(kind, Box::new(handler))
```
jonhoo, line 266: I think we can avoid the boxing here for closures by adding a private type

```rust
#[repr(transparent)]
struct Closure<H>(H);
impl<E, H> JobRunner for Closure<H>
    where H: Fn(Job) -> Result<(), E> + Send + Sync + 'static
{
    type Error = E;
    fn run(&self, job: Job) -> Result<(), Self::Error> {
        (self.0)(job)
    }
}
```

and then passing `Closure(handler)` here

src/consumer/mod.rs
```diff
@@ -52,8 +52,9 @@ pub trait JobRunner: Send + Sync {
     /// A handler function that runs a job.
     fn run(&self, job: Job) -> Result<(), Self::Error>;
 }
+type BoxedJobRunner<E> = Box<dyn JobRunner<Error = E>>;
```
jonhoo, line 55: I don't think we're using this type alias anywhere?

jonhoo (changes requested):
Just one more bit to avoid the box, and then I think we're done!

src/consumer/mod.rs
Kab1r: The type alias is still used in the [unchanged `Consumer` struct](https://github.com/jonhoo/faktory-rs/pull/51/files#diff-c0b949907fe3d15f1fd945390ec35d975eb141f2f4a5724dcd82a9bf7470593aR196) (and it's builder):
```rust
pub struct Consumer<S, E>
where
    S: Read + Write,
{
    c: Client<S>,
    worker_states: Arc<Vec<Mutex<WorkerState>>>,
    callbacks: Arc<FnvHashMap<String, BoxedJobRunner<E>>>,
    terminated: bool,
}
```

jonhoo: Ah, I missed that, thanks!

jonhoo (approved):
Looks good to me — thanks for persisting!

jonhoo:
Released in 0.12.5 :tada:

PR #53, opened 2024-02-04
Add "tls' feature workflow and basic test

rustworthy:
- `docker` directory in the project's root with the Faktory behind NGINX docker compose set-up;
- update the `roundtrip` test to conditionally instantiate non-standard streams for Producer and Consumer;
- add `tls` workflow (the `faktory.local.crt` was self-signed, and the request was for  358000 days, so that the CI would not fail on it);

Why `macos` runner [excluded](https://github.com/jonhoo/faktory-rs/pull/53/commits/3e25fd789b31910fd8da422d29e2da417f8ca22d)?
It's super annoying :( Some runs will go [fine] (https://github.com/jonhoo/faktory-rs/actions/runs/7773048733/job/21196231638), another will fail with network [error](https://github.com/jonhoo/faktory-rs/actions/runs/7773048733/job/21196231732), or even worse - both will fail with the same [connection refused](https://github.com/jonhoo/faktory-rs/actions/runs/7772907678). Also a tweak would be needed in the test like this:

```rust
let connector = if cfg!(target_os = "macos") {
            TlsConnector::builder()
                // Danger! Only for testing!
                // On the macos CI runner, the certs are not trusted:
                // { code: -67843, message: "The certificate was not trusted." }
                .danger_accept_invalid_certs(true)
                .build()
                .unwrap()
        } else { /* ..add the server cert to root certs .. */ }
```
So, not ideal, but staying the `ubuntu-latest` runner with `stable` and `beta` channels tested

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/53)


tests/real/community.rs
```diff
@@ -62,6 +62,77 @@ fn roundtrip() {
...
+        .unwrap();
+    let tls = || {
+        let connector = if cfg!(target_os = "macos") {
+            TlsConnector::builder()
+                // Danger! Only for testing!
+                // On the macos CI runner, the certs are not trusted:
+                // { code: -67843, message: "The certificate was not trusted." }
+                .danger_accept_invalid_certs(true)
+                .build()
+                .unwrap()
```
rustworthy, line 105: Made tests on macos [succeed](https://github.com/jonhoo/faktory-rs/actions/runs/7773140130/job/21196438373). Although the tests are  no longer executed on a macos runner, leaving this dangerous workaround behind for possible future use.  ...and as a reminder of how painful that was :)

docker/compose.yml
```diff
@@ -0,0 +1,16 @@
+services:
+  faktory:
+    image: contribsys/faktory:latest
+    command: "/faktory -b :7419 -w :7420"
+  nginx:
+    depends_on:
+      - faktory
+    image: nginx:1.25-alpine
+    ports:
+      - "127.0.0.1:7419:7419"
```
rustworthy, line 10: This is for the standard integration tests (those not using `TlsStream`, i.e. the absolute majority) to run as usual.

Makefile
```diff
@@ -1,5 +1,6 @@
-FAKTORY_HOST=127.0.0.1
+FAKTORY_HOST=localhost
```
rustworthy, line 1: Should be `localhost` since the ssl certificate is for this common name. Does not affect the tests or local development in any way.

tests/real/community.rs
```diff
@@ -62,6 +62,77 @@ fn roundtrip() {
     assert_eq!(job.args(), &[Value::from("z")]);
 }
 
+#[test]
+#[cfg(feature = "tls")]
+fn roundtrip_tls() {
+    use native_tls::{Certificate, TlsConnector};
+    use std::{env, fs::File, io::Read};
+
+    if env::var_os("FAKTORY_URL").is_none() || env::var_os("FAKTORY_URL_SECURE").is_none() {
```
jonhoo, line 71: I think this should only return early if `FAKTORY_URL_SECURE` is unset, right? If `FAKTORY_URL` is unset, this test can still run.

tests/real/community.rs
```diff
@@ -62,6 +62,77 @@ fn roundtrip() {
...
+        return;
```
jonhoo, line 72: May be worth leaving a comment linking to https://github.com/rust-lang/rust/issues/68007 here for future reference.

tests/real/community.rs
```diff
@@ -62,6 +62,77 @@ fn roundtrip() {
...
+
+    let mut cert = String::new();
+    let cert_path = env::current_dir()
+        .unwrap()
+        .join("docker")
+        .join("certs")
+        .join("faktory.local.crt");
+    File::open(cert_path)
+        .and_then(|mut f| f.read_to_string(&mut cert))
+        .unwrap();
```
jonhoo, line 96:
```suggestion
    let cert = std::fs::read_to_string(cert_path).unwrap();
```

tests/real/community.rs
jonhoo: Given this is all testing in CI anyway, I'd also be fine just running the entire test (on all OSes) with `danger_accept_invalid_certs`. But leaving it here with docs like you have seems fine :+1:

tests/real/community.rs
```diff
@@ -62,6 +62,77 @@ fn roundtrip() {
...
+                .build()
+                .unwrap()
+        };
+        TlsStream::with_connector(
+            connector,
+            Some(
+                std::env::var_os("FAKTORY_URL_SECURE")
+                    .unwrap()
+                    .to_str()
+                    .unwrap(),
```
jonhoo, line 119: Why not
```suggestion
                std::env::var("FAKTORY_URL_SECURE").unwrap()
```
?

.github/workflows/tls.yml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: ubuntu-latest / ${{ matrix.toolchain }} / tls
+    strategy:
+      fail-fast: false
+      matrix:
+        # run on stable and beta to ensure that tests won't break on the next version 
+        # of the rust toolchain
+        toolchain: [stable, beta]
```
jonhoo, line 23: I don't think we need both stable and beta for this test. Just stable is fine.

.github/workflows/tls.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          toolchain: ${{ matrix.toolchain }}
+      - name: cargo generate-lockfile
+        if: hashFiles('Cargo.lock') == ''
+        run: cargo generate-lockfile
+      - name: Run tests
+        env:
+          FAKTORY_URL: tcp://localhost:7419
+          FAKTORY_URL_SECURE: tcp://localhost:17419
+        run: cargo test --locked --features tls --all-targets
```
jonhoo, line 41: To speed up CI, I would actually prefer for this job to _only_ run the TLS tests. Which suggests they should live in their own file in `tests/` so that you can pass `--test tls` here.

jonhoo (changes requested):
Nice!

tests/real/community.rs
rustworthy: done

tests/real/community.rs
rustworthy: Indeed, fixed

tests/real/community.rs
rustworthy: No good reason. Fixed

tests/real/community.rs
rustworthy: Updated. I guess should have first applied the suggestions and then make restructures the `/tests` dir.

docker/compose.yml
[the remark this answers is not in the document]
rustworthy: Yeah, I remember I was check it with `cargo package --list --allow-dirty` when adding a `Makefile` , but there I noticed `REAMDE.tpl`, `.github` and `tests` had been already packaged, which I thought was an adopted policy - package everything the project comprises of so that in case of disaster one could restore literally everything.

I'll exclude `docker` and `.github`, but the tests should stay I believe? Found an accepted [answer](https://users.rust-lang.org/t/should-test-files-and-fixtures-be-excluded-from-cargo-projects-when-packaging/81949/2):
```txt
...people can run your tests in the version you published to crates.io (as opposed to having to cross-reference with your repository commits).
```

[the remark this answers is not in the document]
rustworthy: image

docker/compose.yml
```diff
@@ -0,0 +1,16 @@
+services:
+  faktory:
+    image: contribsys/faktory:latest
+    command: "/faktory -b :7419 -w :7420"
+  nginx:
+    depends_on:
+      - faktory
+    image: nginx:1.25-alpine
```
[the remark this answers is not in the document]
rustworthy: Done. It's possible with the help of a little trick as described [here](https://github.com/dependabot/dependabot-core/issues/390#issuecomment-1062170379).

Makefile
```diff
@@ -26,8 +27,13 @@ faktory:
 faktory/kill:
 	docker stop faktory
 
-README.md: README.tpl src/lib.rs
-	cargo readme > README.md
+.PHONY: faktory/tls
+faktory/tls:
+	docker compose -f docker/compose.yml up -d --build
```
rustworthy, line 32: `--build` flag needed otherwise a cached image will be used, even if we update the NGINX or Faktory images versions in their corresponding dockerfiles. I am also adding this flag on the CI, though of course no images are cached since we are starting a-fresh each time. But still.

jonhoo (approved):
Excellent, looks all good to me!

jonhoo:
Released in 0.12.5 :tada:

PR #54, opened 2024-02-07
Bulk Push Jobs

rustworthy:
- adding `enqueue_many` method on the `Producer` (internally uses `PUSHB` Faktory command);
 - adding `make sort` command and CI job to address the [thread](https://github.com/jonhoo/faktory-rs/pull/48#discussion_r1477280308);
 - running `make sort` against the sources;

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/54)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -40,7 +40,7 @@ jobs:
   # See: https://github.com/rust-lang/rustfmt/issues/5083
   imports:
     runs-on: ubuntu-latest
-    name: nightly / fmt (imports)
+    name: nightly / fmt (import grouping)
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -34,6 +34,23 @@ jobs:
           components: rustfmt
       - name: cargo fmt --check
         run: cargo fmt --check
+  # This is currently a dedicated job due to the rustfmt's `group_imports` configuration
+  # option being available on the nightly channel only as of February 2024.
+  # Once stabilized, can be merged with the `stable / fmt` job in this workflow.
+  # See: https://github.com/rust-lang/rustfmt/issues/5083
+  imports:
+    runs-on: ubuntu-latest
+    name: nightly / fmt (imports)
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Install nightly
+        uses: dtolnay/rust-toolchain@nightly
+        with:
+          components: rustfmt
+      - name: cargo +nightly fmt -- --config group_imports=one --check
+        run: cargo +nightly fmt -- --config group_imports=one --check
   clippy:
     runs-on: ubuntu-latest
     name: ${{ matrix.toolchain }} / clippy
--- a/src/producer/mod.rs
+++ b/src/producer/mod.rs
@@ -130,7 +130,7 @@ impl<S: Read + Write> Producer<S> {
         let jobs_count = jobs.len();
         let errors: HashMap<String, String> = self
             .c
-            .issue(&PushBulk::from(jobs))?
+            .issue(&PushBulk::from(jobs.collect::<Vec<_>>()))?
             .read_json()?
             .expect("Faktory server sends {} literal when there are no errors");
         if errors.is_empty() {
--- a/src/producer/mod.rs
+++ b/src/producer/mod.rs
@@ -122,20 +122,21 @@ impl<S: Read + Write> Producer<S> {
         &mut self,
         jobs: J,
     ) -> Result<(usize, Option<HashMap<String, String>>), Error>
-      where J: IntoIterator<Item = Job>,
-            J::IntoIter: ExactSizeIterator
+    where
+        J: IntoIterator<Item = Job>,
+        J::IntoIter: ExactSizeIterator,
     {
         let jobs = jobs.into_iter();
-        let njobs = jobs.len();
+        let jobs_count = jobs.len();
         let errors: HashMap<String, String> = self
             .c
-            .issue(&PushBulk::from(jobs))?
+            .issue(&PushBulk::from(jobs.collect::<Vec<_>>()))?
             .read_json()?
             .expect("Faktory server sends {} literal when there are no errors");
         if errors.is_empty() {
-            return Ok((jobs.len(), None));
+            return Ok((jobs_count, None));
         }
-        Ok((jobs.len() - errors.len(), Some(errors)))
+        Ok((jobs_count - errors.len(), Some(errors)))
     }
 
     /// Retrieve information about the running server.
--- a/src/producer/mod.rs
+++ b/src/producer/mod.rs
@@ -1,5 +1,6 @@
 use crate::error::Error;
-use crate::proto::{self, Client, Info, Job, Push, QueueAction, QueueControl};
+use crate::proto::{self, Client, Info, Job, Push, PushBulk, QueueAction, QueueControl};
+use std::collections::HashMap;
 use std::io::prelude::*;
 use std::net::TcpStream;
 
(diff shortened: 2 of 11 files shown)


.github/workflows/check.yml
```diff
@@ -34,6 +34,23 @@ jobs:
           components: rustfmt
       - name: cargo fmt --check
         run: cargo fmt --check
+  # This is currently a dedicated job due to the rustfmt's `group_imports` configuration
+  # option being available on the nightly channel only as of February 2024.
+  # Once stabilized, can be merged with the `stable / fmt` job in this workflow.
+  # See: https://github.com/rust-lang/rustfmt/issues/5083
+  imports:
+    runs-on: ubuntu-latest
+    name: nightly / fmt (imports)
```
jonhoo, line 43:
```suggestion
    name: nightly / fmt (import grouping)
```

tests/real/community.rs
```diff
@@ -164,6 +164,87 @@ fn queue() {
...
+        .unwrap();
+    assert_eq!(enqueued_count, 3);
+    assert!(errors.is_none()); // error-free
+
+    // From the Faktory source code, we know that:
+    // 1) job ID should be at least 8 chars long string; NB! Should be taken in account when introducing `Jid` new type;
+    // 2) jobtype should be a non-empty string;
+    // 3) job cannot be reserved for more than 86400 days;
+    // ref: https://github.com/contribsys/faktory/blob/main/manager/manager.go#L192-L203
+    // Let's break these rules:
```
jonhoo, line 192: Nice!

src/producer/mod.rs
```diff
@@ -106,6 +107,32 @@ impl<S: Read + Write> Producer<S> {
...
+    /// The returned `Ok` result will contain a tuple of enqueued jobs count and an option of a hash map
+    /// with job ids mapped onto error messages. Therefore `Ok(n, None)` will indicate that all n jobs
+    /// have been enqueued without errors.
+    ///
+    /// Note that this is not an all-or-nothing operation: jobs that contain errors will not be enqueued,
+    /// while those that are error-free _will_ be enqueued by the Faktory server.
+    pub fn enqueue_many(
+        &mut self,
+        jobs: &[Job],
+    ) -> Result<(usize, Option<HashMap<String, String>>), Error> {
```
jonhoo, line 124: What do you think of doing something like this?
```suggestion
    pub fn enqueue_many<J>(
        &mut self,
        jobs: J,
    ) -> Result<(usize, Option<HashMap<String, String>>), Error>
      where J: IntoIterator<Item = Job>,
            J::IntoIter: ExactSizeIterator
    {
        let jobs = jobs.into_iter();
        let njobs = jobs.len();
```
You'd also want to update `PushBulk` accordingly (although it'll only need an `Iterator<Item = Job>` bound).

src/proto/single/cmd.rs
```diff
@@ -245,6 +244,31 @@ impl FaktoryCommand for Push {
 
 // ----------------------------------------------
 
+pub struct PushBulk<'a>(&'a [Job]);
```
jonhoo, line 247: Feels weird for this to hold references to `Job` when `Push` holds an owned `Job`. See the suggestion on `enqueue_many` for one alternative.

In the next major version we may want to make both this and `Push` refer to `&Job` (or maybe `Cow<'_, Job>`?) so that callers don't _have_ to pass in owned `Job` given that `issue` doesn't require it.

jonhoo (changes requested):
Formatting changes look good! One suggestion on the `enqueue_many` side of things.

src/producer/mod.rs
rustworthy: Please see the updated version, as for the `PushBulk` - updating it expect a vector of jobs. I guess I might be missing a way to collect an iterator behind a shared refence in `issue`, while `Serialize` is not automatically implemented for the `Iterator<Item = S> where S: Serialize`

src/proto/single/cmd.rs
rustworthy: As of now we do not need to consume a job. And I was about to say there is not a single place where we need more than read-only access to the job, once is has been built by the end-user, BUT there will be: in the Ent Faktory batches:
```rust
pub fn add(&mut self, mut job: Job) -> Result<Option<serde_json::Value>, Error> {
        let bid = job.custom.insert("bid".into(), self.bid.clone().into());
        self.prod.enqueue(job).map(|_| bid)
}
```
In this case we will need a mutable reference and we will be mutating the job returning the user the previous value of `bid` if there was one - so that it's not a surprise for them. So `&mut Job` would be just enough.

As for the `Cow` - this will require `Job` to be `Clone`, while it's not. My understanding was that you did not make it `Clone` initially for a reason: this would clone the `jid` and potentially can lead to duplicates, since the user may think they can simply clone a job - if they need a couple of them - and send it to the server. On the other hand, implementing `Clone` where `jid` will be re-generated - may be surprising behavior which we want to avoid in all cases. Was this assumption correct?

So, I think we can go with `& Job` and `&mut Job` in the future.

tests/real/community.rs
```diff
@@ -175,7 +175,7 @@ fn test_jobs_pushed_in_bulk() {
 
     let mut p = Producer::connect(None).unwrap();
     let (enqueued_count, errors) = p
-        .enqueue_many(&[
+        .enqueue_many(vec![
```
jonhoo, line 178: My hope is that you can now use `[]` here (so not `vec![]` _or_ `&[]`).

src/producer/mod.rs
```diff
@@ -122,20 +122,21 @@ impl<S: Read + Write> Producer<S> {
...
+        J: IntoIterator<Item = Job>,
+        J::IntoIter: ExactSizeIterator,
     {
         let jobs = jobs.into_iter();
-        let njobs = jobs.len();
+        let jobs_count = jobs.len();
         let errors: HashMap<String, String> = self
             .c
-            .issue(&PushBulk::from(jobs))?
+            .issue(&PushBulk::from(jobs.collect::<Vec<_>>()))?
```
jonhoo, line 133: I'm hoping we can maybe avoid the `collect` here too. The trick is going to be to also make `PushJobs` generic (see comment there).

src/proto/single/cmd.rs
```diff
@@ -244,22 +244,22 @@ impl FaktoryCommand for Push {
 
 // ----------------------------------------------
 
-pub struct PushBulk<'a>(&'a [Job]);
+pub struct PushBulk(Vec<Job>);
 
-impl<'a> Deref for PushBulk<'a> {
-    type Target = &'a [Job];
+impl Deref for PushBulk {
```
jonhoo, line 249: We'll probably want to remove this and instead implement `Serialize` for this method, which we can do with something like:
```rust
impl<I> Serialize for PushBulk<I>
where
    I: ExactSizeIterator<Item = Job>,
{
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        let mut seq = serializer.serialize_seq(Some(self.0.len()))?;
        for e in &mut self.0 {
            seq.serialize_element(e)?;
        }
        seq.end()
    }
}
```

rustworthy: I only managed to do with interior mutability. Please have a look once again.

jonhoo: Ahhh, it's because `FaktoryCommand` takes `&self`... That's annoying. The `RefCell` approach won't work either if `PushBulk::issue` is ever called more than once. And `FaktoryCommand::issue` takes `&self` precisely to prevent excessive cloning.

Hmm hmm hmm. We _could_ change `FaktoryCommand::issue` to take `self` (no `&`), though that feels excessive. There may be a way to do it with an awful bound like `for<'a> &'a I: IntoIterator`, but I think probably just stick with the version you had where `PushBulk` holds a `Vec<Job>` for now. Sorry for the back and forth!

rustworthy: No worries, reverted.

jonhoo (approved):
Thanks!

jonhoo:
Made one small adjustment after the fact so that we don't have to guarantee in the future that we continue to hold a `Vec`:
https://github.com/jonhoo/faktory-rs/commit/0bd2ab0dd1603e972f1bb5ba02669f9ab684fd8c

jonhoo:
Released in 0.12.5 :tada:

PR #57, opened 2024-03-21
Handle SIGTERM and signals from user space

rustworthy:
As per #4

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/57)


Cargo.toml
```diff
@@ -41,14 +41,17 @@ tokio = { version = "1.35.1", features = [
     "net",
     "time",
     "macros",
+    "signal",
     "rt-multi-thread",
 ] }
 tokio-rustls = { version = "0.25.0", optional = true }
+tracing = "0.1"
```
rustworthy, line 48: @jonhoo I saw them putting it under feature flag and combining with `log`, e.g. [`here`](https://github.com/fede1024/rust-rdkafka/blob/master/src/log.rs). But my understanding is that this has grown into a standard and can be included without opt-in. Those subscribing will be able to opt-out of the events originated from the crate free of CPU/mem charge, wdyt?

src/worker/mod.rs
```diff
@@ -380,7 +406,7 @@ impl<
     where
         Q: AsRef<str>,
     {
-        use std::process;
+        self.forever = true;
```
rustworthy, line 409: this is to be able to switch off the corresponding `select!` branch for normal (returning) `run` (as opposed to `run_to_completion`)

tests/real/community.rs
```diff
@@ -310,3 +310,85 @@ async fn test_jobs_created_with_builder() {
...
+            let complexity = j.args().iter().next().unwrap().as_u64().unwrap();
+            sender.send(true).await.unwrap(); // inform that we are now starting to process the job
+            tokio::time::sleep(tokio::time::Duration::from_millis(complexity)).await;
+            Ok::<(), io::Error>(())
+        })
+    });
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_shutdown_signals_handling() {
```
rustworthy, line 336: Ctrl-C signal have been tested with the help of `run_to_completion` example

src/worker/channel.rs
```diff
@@ -0,0 +1,34 @@
+#[cfg(doc)]
+use super::{Worker, WorkerBuilder};
+
+use tokio::sync::mpsc::{self, Receiver, Sender};
+
+/// Message sent to running worker.
+///
+/// See documentation to [`Worker::run`](Worker::run)
+pub enum Message {
```
rustworthy, line 9: this might me extended in the future, so I think it's not exhaustove

jonhoo:
just fyi, I've seen this, but am holding off on reviewing until we get #49 over the finish line.

src/worker/builder.rs
```diff
@@ -106,6 +121,30 @@ impl<E: 'static> WorkerBuilder<E> {
...
+    /// [docs](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) for
+    /// how to set the upper limit on the number of threads in the mentioned pool and other details.
+    ///
+    /// Note that it is not recommended to mix blocking and non-blocking tasks and so if many of your
+    /// handlers are blocking, you will want to launch a dedicated worker process (at least a separate
+    /// Tokio Runtime) where only blocking handlers will be used.
+    ///
+    /// Also note that only one single handler per job kind is supported. Registering another handler
+    /// for the same job kind will silently override the handler registered previously.
+    pub fn register_blocking_fn<K, H>(mut self, kind: K, handler: H) -> Self
```
rustworthy, line 138: Rationale:

We probably should not be discriminating those who have been using the crate for CPU intensive tasks running on the background, e.g. image processing.

jonhoo:
I'd prefer that we don't put signal handling directly into this library, and instead do something like what `axum` does and allow the [user to supply a shutdown future](https://docs.rs/axum/0.7/axum/serve/struct.Serve.html#method.with_graceful_shutdown).

jonhoo:
Also, would you mind doing the examples and tracing in a separate PR just to make it easier to review and land this piecemeal?

rustworthy:
> Also, would you mind doing the examples and tracing in a separate PR just to make it easier to review and land this piecemeal?

Sure. It's not that I like mixing unrelated stuff one PR, rather the example and tracing became a side-effect of adding a signal handling logic, namely testing ctrl c handling

src/worker/builder.rs
rustworthy: this  has been moved :smile:

rustworthy: now owned by #65

src/lib.rs
```diff
@@ -48,9 +48,9 @@
 //!
 //! ```no_run
 //! # tokio_test::block_on(async {
-//! use faktory::WorkerBuilder;
+//! use faktory::Worker;
 //! use std::io;
-//! let mut w = WorkerBuilder::default()
+//! let mut w = Worker::builder()
```
jonhoo, line 53: Should this go in #64?

src/worker/builder.rs
```diff
@@ -7,13 +7,17 @@ use std::future::Future;
 use tokio::io::{AsyncRead, AsyncWrite, BufStream};
 use tokio::net::TcpStream as TokioStream;
 
+pub(crate) const GRACEFUL_SHUTDOWN_PERIOD_MILLIS: u64 = 5_000;
```
jonhoo, line 10:
```suggestion
pub(crate) const GRACEFUL_SHUTDOWN_PERIOD: Duration = Duration::from_secs(5);
```

src/worker/builder.rs
```diff
@@ -7,13 +7,17 @@ use std::future::Future;
...
+
 /// Convenience wrapper for building a Faktory worker.
 ///
 /// See the [`Worker`] documentation for details.
 pub struct WorkerBuilder<E> {
     opts: ClientOptions,
     workers_count: usize,
     callbacks: CallbacksRegistry<E>,
+    shutdown_timeout: u64,
```
jonhoo, line 19: Let's make this be a `Duration` and extract the millis when needed.

src/worker/builder.rs
```diff
@@ -90,6 +96,65 @@ impl<E: 'static> WorkerBuilder<E> {
...
+        self.shutdown_signal = Some(Box::pin(signal));
+        self
+    }
+
+    /// Set the graceful shutdown period in milliseconds. Defaults to 5000.
+    ///
+    /// This will be used once the worker is sent a termination signal whether it is at the application
+    /// (via a signalling future, see [`WorkerBuilder::with_graceful_shutdown`]) or OS level (via Ctrl-C signal,
+    /// see [`Worker::run_to_completion`]).
+    pub fn graceful_shutdown_period(mut self, millis: u64) -> Self {
```
jonhoo, line 153: This should take a `Duration`.

Also, should we allow users to _not_ set a shutdown period (i.e., to never have shutdown aborted)?

src/worker/mod.rs
```diff
@@ -148,6 +154,9 @@ pub struct Worker<S: AsyncWrite + Send + Unpin, E> {
     worker_states: Arc<state::WorkerStatesRegistry>,
     callbacks: Arc<CallbacksRegistry<E>>,
     terminated: bool,
+    forever: bool,
+    shutdown_timeout: u64,
```
jonhoo, line 158:
```suggestion
    shutdown_timeout: Duration,
```

src/worker/mod.rs
```diff
@@ -337,10 +358,14 @@ impl<
...
-    /// until the server tells the worker to disengage (`Ok` is returned).
+    /// Run this worker on the given `queues`.
     ///
-    /// The value in an `Ok` indicates the number of workers that may still be processing jobs.
+    /// Will run the worker until an I/O error occurs (`Err` is returned), or until the server tells the worker
+    /// to disengage (`Ok` is returned), or a signal from the user-space code has been received via a future
+    /// supplied to [`WorkerBuilder::with_graceful_shutdown`](`Ok` is returned).
+    ///
+    /// The value in an `Ok` indicates the number of workers that may still be processing jobs, but `0` can also
+    /// indicate the [graceful shutdown period](WorkerBuilder::graceful_shutdown_period) has been exceeded.
```
jonhoo, line 368: Maybe we should introduce an `enum` here to tell the user which `Ok`-returning case they are in?

src/worker/mod.rs
```diff
@@ -369,45 +394,83 @@ impl<
...
+        };
 
-        // we want to expose worker errors, or otherwise the heartbeat error
-        let mut results = Vec::with_capacity(workers_count);
-        while let Some(res) = join_set.join_next().await {
-            results.push(res.expect("joined ok"));
-        }
+        let report = tokio::select! {
+            // A signal SIGTERM from the OS received.
+            _ = tokio::signal::ctrl_c(), if self.forever => {
```
jonhoo, line 408: I still don't want us to have any signal handling code inside the core library. Instead, we should make it possible to configure `run_to_completion` with a graceful shutdown feature as well.

src/worker/mod.rs
```diff
@@ -369,45 +394,83 @@ impl<
...
+                    },
+                    nrunning = self.force_fail_all_workers("termination signal received over channel") => {
+                        nrunning
+                    }
+                };
+                self.terminated = true;
+                Ok(nrunning)
+            },
+           // A signal from the Faktory server received or an error occurred.
+           exit = self.listen_for_heartbeats(&statuses) => {
```
jonhoo, line 435: Since this isn't cancellation-safe, we should document why we still use it in a `select!` and expect it to be okay. I think it's because if we do cancel `listen_for_heartbeats` we are exiting anyway, and the connection will never be used again?

src/worker/mod.rs
```diff
@@ -2,11 +2,16 @@ use super::proto::{Client, Reconnect};
...
+use std::future::Future;
+use std::pin::Pin;
+use std::process;
 use std::sync::{atomic, Arc};
 use std::{error::Error as StdError, sync::atomic::AtomicUsize};
 use tokio::io::{AsyncBufRead, AsyncWrite};
 use tokio::net::TcpStream;
 use tokio::task::{AbortHandle, JoinSet};
+use tokio::time::sleep as tokio_sleep;
+use tokio::time::Duration as TokioDuration;
```
jonhoo, line 14: I think this is just a re-export of `std::time::Duration`?

src/worker/mod.rs
```diff
@@ -369,45 +394,83 @@ impl<
...
+                        process::exit(0);
+                    }
+                }
+            },
+            // A signal from the user space received.
+            _ = async { let signal = cancel_signal.unwrap(); signal.await }, if cancel_signal.is_some() => {
+                let nrunning = tokio::select! {
+                    _ = tokio_sleep(TokioDuration::from_millis(self.shutdown_timeout)) => {
+                        0
+                    },
```
jonhoo, line 426: Instead of this you can just use `tokio::time::timeout` to wrap the future from `force_fail_all_workers`.

tests/real/community.rs
```diff
@@ -340,3 +341,75 @@ async fn test_jobs_created_with_builder() {
...
+) -> Box<
+    dyn Fn(Job) -> Pin<Box<dyn Future<Output = Result<(), io::Error>> + Send>>
+        + Send
+        + Sync
+        + 'static,
+> {
+    return Box::new(move |j: Job| {
+        let sender = sync::Arc::clone(&sender);
+        Box::pin(async move {
+            let complexity = j.args().iter().next().unwrap().as_u64().unwrap();
```
jonhoo, line 359: Why not just `j.args()[0]`?

jonhoo (changes requested):
Sorry it took me a while (again) :sweat_smile:

rustworthy:
> Sorry it took me a while (again) 😅

No worries, it actually takes me that exact Duration to iterate over the things

src/worker/builder.rs
rustworthy: Done!

src/worker/builder.rs
rustworthy: Updated.

My understanding is when dealing with futures or long-running tasks in general you almost always want to have a timeout. It's just that many (most?) of the operations we usually deal with are quick enough to finish for us to not get into a "hanging" state.

At the same time the question is where did I get the default value from (and the answer may not sound convincing). I also just looked up the `reqwest` API and they are defaulting the [timeout](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout) to `None` .

Yes, I think it should be defaulted to `None` and should be called simply `shutdown_timeout` (even if it's being performed with disgrace :smile: ) following the `reqwest`'s namings.

src/lib.rs
rustworthy: Moved to #64

src/worker/mod.rs
rustworthy: Intrroduced `RunCeaseReason` and updated the docs and the tests

src/worker/mod.rs
rustworthy: Indeed :facepalm:

tests/real/community.rs
rustworthy: I am already doing this automatically, but I agree that since we are not handling `None` properly, direct indexing makes more sence

src/worker/mod.rs
rustworthy: Exactly! Even though it  is not cancellation safe, we are ok, because we are not selecting in a loop and we are eventually _either_ marking the worker as terminated (this is being  verified as soon as we enter Worker::run, _or_ re-creating a connection (instantiating a new stream, running a handshake), i.e. we never end up re-using the "broken" connection.

Added docs.

src/worker/mod.rs
rustworthy: Removed this select arm and also the tokio's `signal` feature from the prod deps.

I guess the reason why we are deciding against ctrl_c handling is the possible conflicts from the user space code event handling. They should (at least will want to) define it in their binary.

rustworthy:
@jonhoo Please have another look.

As for the conflicts, #68 should unblock the pipeline in `main`

jonhoo:
Just FYI — I'm away on holiday this week, so won't get to this until the weekend after next probably 😅

rustworthy:
> Just FYI — I'm away on holiday this week, so won't get to this until the weekend after next probably 😅

I wish you a great holiday! As for the PRs, I still need some extra time for the tech debt one, so this is actually great.

rustworthy:
@jonhoo 
Sorry for pushing amid review request - just pulled the latest from main resolving some minor conflicts

src/worker/mod.rs
```diff
@@ -286,6 +304,23 @@ impl<S: AsyncBufRead + AsyncWrite + Send + Unpin, E: StdError + 'static + Send>
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+/// A reason why [`Worker::run`] has discontinued.
+pub enum RunCeaseReason {
```
jonhoo, line 309:
```suggestion
#[non_exhaustive]
pub enum StopReason {
```
As it feels likely that there may be more stop reasons in the future. Also, I think it's unlikely that callers will benefit from exhaustive matching on this `enum`.

src/worker/mod.rs
```diff
@@ -286,6 +304,23 @@ impl<S: AsyncBufRead + AsyncWrite + Send + Unpin, E: StdError + 'static + Send>
...
+    /// A signal from the user space received.
+    ///
+    /// A future provided via [`WorkerBuilder::with_graceful_shutdown`] has resolved
+    /// signalling the worker to stop.
+    CancelSignal,
```
jonhoo, line 314:
```suggestion
    /// Graceful shutdown completed.
    ///
    /// A future provided via [`WorkerBuilder::with_graceful_shutdown`] has resolved
    /// signalling the worker to stop.
    GracefulShutdown,
```

src/worker/mod.rs
```diff
@@ -286,6 +304,23 @@ impl<S: AsyncBufRead + AsyncWrite + Send + Unpin, E: StdError + 'static + Send>
...
+
+    /// A signal from the Faktory server received.
+    ///
+    /// Under the hood, the worker is being in constant communication with the Faktory server,
+    /// not only fetching jobs and reporting on processing results, but also listening for
+    /// the server's instructions, one of which can be to disengage.
+    FaktoryInstruction,
```
jonhoo, line 321:
```suggestion
    /// The Faktory server asked us to shut down.
    ///
    /// Under the hood, the worker is being in constant communication with the Faktory server,
    /// not only fetching jobs and reporting on processing results, but also listening for
    /// the server's instructions, one of which can be to disengage (e.g., to indicate that the
    /// server is shutting down.
    ServerInstruction,
```

src/worker/mod.rs
```diff
@@ -369,51 +412,86 @@ impl<
...
+        // Either we have already taken this future and are polling it elsewhere (see Worker::run_to_completion),
+        // or they've never provided it to us (reminder: they do so via WorkerBuilder::with_graceful_shutdown)
+        let cancel_signal = self.shutdown_signal.take();
+
+        let report = tokio::select! {
+            // A signal from the user space received.
+            _ = async { let signal = cancel_signal.unwrap(); signal.await }, if cancel_signal.is_some() => {
+                let nrunning = tokio::select! {
+                     _ = tokio_sleep(self.shutdown_timeout.unwrap()), if self.shutdown_timeout.is_some() => {
+                        0
```
jonhoo, line 424: A future TODO: it would be nice if we could give an estimate of how many workers actually remain here, but I think this is fine for now.

src/worker/mod.rs
```diff
@@ -369,51 +412,86 @@ impl<
...
+                Ok((RunCeaseReason::CancelSignal, nrunning))
+            },
+            // A signal from the Faktory server received or an error occurred.
+            // Even though `Worker::listen_for_hearbeats` is not cancellation safe, we are ok using it here,
+            // since we are not `select!`ing in a loop and we are eventually _either_ marking the worker as
+            // terminated, _or_ re-creating a connection, i.e. we never end up re-using the "broken" connection.
+            exit = self.listen_for_heartbeats(&statuses) => {
+                // there are a couple of cases here:
+                //  - we got TERMINATE, so we should just return, even if a worker is still running
+                //  - we got TERMINATE and all workers has exited
```
jonhoo, line 440:
```suggestion
                //  - we got TERMINATE and all workers have exited
```

src/worker/mod.rs
```diff
@@ -369,51 +412,86 @@ impl<
                 .await?;
         }
 
-        let exit = self.listen_for_heartbeats(&statuses).await;
+        // Either we have already taken this future and are polling it elsewhere (see Worker::run_to_completion),
+        // or they've never provided it to us (reminder: they do so via WorkerBuilder::with_graceful_shutdown)
+        let cancel_signal = self.shutdown_signal.take();
```
jonhoo, line 417: I'm worried about taking the shutdown signal here, because it means that if the user decides to reconnect using the same `Worker`, their graceful shutdown future is no longer there and active. I think we should only `take` it if it resolves (because then it's _definitely_ not useful again in the future). And then we should also document (on `with_graceful_shutdown`) that if the future resolves, it is then removed and no longer taken into account if the `Worker` _is_ reused.

src/worker/mod.rs
```diff
@@ -369,51 +412,86 @@ impl<
...
         Q: AsRef<str>,
     {
-        use std::process;
-        while self.run(queues).await.is_err() {
-            if self.reconnect().await.is_err() {
-                break;
-            }
-        }
+        let sig = self.shutdown_signal.take();
+        tokio::select! {
```
jonhoo, line 477: Why can't this just call `self.run` and allow `run` to deal with the graceful shutdown logic? Especially if we make the change around `.take()` above that I suggested.

src/worker/mod.rs
```diff
@@ -337,15 +375,20 @@ impl<
...
-    /// The value in an `Ok` indicates the number of workers that may still be processing jobs.
+    /// The value in an `Ok` holds a tuple with the reason why the run has discontinued (see [`RunCeaseReason`])
+    /// and the number of workers that may still be processing jobs. Note that `0` can also indicate that
+    /// the [graceful shutdown period](WorkerBuilder::shutdown_timeout) has been exceeded.
     ///
     /// If an error occurred while reporting a job success or failure, the result will be re-reported to the server
     /// without re-executing the job. If the worker was terminated (i.e., `run` returns  with an `Ok` response),
     /// the worker should **not** try to resume by calling `run` again. This will cause a panic.
-    pub async fn run<Q>(&mut self, queues: &[Q]) -> Result<usize, Error>
+    pub async fn run<Q>(&mut self, queues: &[Q]) -> Result<(RunCeaseReason, usize), Error>
```
jonhoo, line 391: How do you feel about returning a `struct` here with `pub` fields rather than a tuple?

jonhoo (changes requested):
This is nearly there now, mostly smaller nits left :tada:

src/worker/mod.rs
rustworthy: Hmm I was wondering how this got into the PR, and I remember now that it used to be [`has`](https://github.com/jonhoo/faktory-rs/blob/8c06cc7d0d84d337f2c29da3fab9708cffea143f/src/consumer/mod.rs#L505) and then we corrected it in the make-everything-async merge to be `have`, but I it somehow remained in the initial edition in this branch and so it looks like I am suggesting this, which technically I am :smile: 

I wonder though how you noticed this one among all this whitespace noise :wink:

src/worker/mod.rs
rustworthy: added to the issues (#76)

src/worker/builder.rs
```diff
@@ -90,6 +95,67 @@ impl<E: 'static> WorkerBuilder<E> {
...
+    ///     .register_fn("job_type", move |_| async { Ok::<(), std::io::Error>(()) })
+    ///     .connect(None)
+    ///     .await
+    ///     .unwrap();
+    ///
+    /// // start consuming
+    /// let jh = tokio::spawn(async move { w.run(&["default"]).await });
+    ///
+    /// // send a signal to eventually return control (upon graceful shutdown)
+    /// token.cancel();
```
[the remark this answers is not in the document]
rustworthy: Added in here and - since it's a no_run - added the same check for `!jh.is_finished()` to the test

src/worker/mod.rs
rustworthy: Introduced a `StopDetails` that holds the reason and nrunning and put it to a dedicated module.
Do you think we should  keep this stop details container non_exhaustive as well?

And yes - `run cease` is  really nothing but `stop`. :) The idea was to prefix this guy with the method name, i.e. run - run-cease.

src/worker/mod.rs
rustworthy: You are right, there is one scenario that I've missed: if we get an error from heartbeat(), the worker will _not_ be marked as `terminated` and so they may try to call `run` again. In all other cases the worker is marked as terminated and running this work will panic at the check we are performing. So in all other cases they will need to construct a new worker and run it again.

As for taking in case the future resolves, I _think_ there is no need to do so, since if it resolves we are getting into the block marking the worker as terminated. Added to the docs for `WorkerBuilder::with_graceful_shutdown` and `Worker::run`

Also updated this method to not take the signal.

src/worker/mod.rs
rustworthy: We really could not with `take()`, but it's now possible.

rustworthy:
@jonhoo I think all the threads have been addressed.

I thing that I have just realized is that we do not allow to `Worker::run` a terminated worker, and we drop the processing workers we have spawned, _but_  `Worker::run_one` is still possible since the coordinating worker is not dropped. 

Imagine the situation where the server told us to stop the consumption (we get this message in the heartbeat task on the coordinator), and we do so and mark the coordinator as terminated. What they can do then is take this worker and call `run_one` in a loop mimicking this way the `Worker::run` logic, but without HEARTBEATing and - most importantly - breaking the server instruction we've received previously. I _think_ this is not something we want to be possible, wdyt?

src/worker/mod.rs
jonhoo: It is sad, though I also don't have a great alternative :sweat_smile: After all, if we instead have an `Other` variant, that'll more or less be equivalent to matching on `_` except that you're given the string name of the other variant (which you could otherwise get by invoking `Debug` or we could add a method for it). I suppose having an `Other` variant would also mean that if a new "officially supported" variant comes along, you'd be told your `match` is now missing an arm. However, that's also the source of the problem, namely that it'd require a breaking change and thus a new major version (which users then need to remember to adopt).

src/worker/stop.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// A future provided via [`WorkerBuilder::with_graceful_shutdown`] has resolved
+    /// signalling the worker to stop.
+    GracefulShutdown,
+
+    /// The Faktory server asked us to shut down.
+    ///
+    /// Under the hood, the worker is being in constant communication with the Faktory server,
+    /// not only fetching jobs and reporting on processing results, but also listening for
+    /// the server's instructions, one of which can be to disengage (e.g., to indicate that the
+    /// server is shutting down.
```
jonhoo, line 19:
```suggestion
    /// Under the hood, the worker is in constant communication with the Faktory server.
    /// It fetches jobs and reports on processing results, but also listens for the server's
    /// instructions. One of those instructions can be to disengage (e.g., to indicate that the
    /// server is shutting down.
```

src/worker/stop.rs
```diff
@@ -0,0 +1,38 @@
...
+    ServerInstruction,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+/// Holds some details aroung a worker's run stoppage, such as the reason why this worker discontinued
+/// and the number of workers that might still be processing jobs at that instant.
+pub struct StopDetails {
+    /// A [`reason`](StopReason) why the worker's run has discontinued.
```
jonhoo, line 27:
```suggestion
    /// The [reason](StopReason) why the worker's run has discontinued.
```

src/worker/stop.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub reason: StopReason,
+
+    /// The number of workers that might still be processing jobs.
+    pub nrunning: usize,
```
jonhoo, line 31:
```suggestion
    pub workers_still_running: usize,
```
better?

src/worker/mod.rs
jonhoo: Yeah, I think that should probably also be `#[non_exhaustive]`. The main downside to doing so is that people can't easily construct their own for testing purposes, but if we have it also `derive(Default)` then maybe that helps with that?

src/worker/mod.rs
jonhoo: Hmm, but now the code has an `Arc<Mutex` instead, which feels unnecessary. I _think_ you should be able to keep it just `ShutdownSignal`. The part that's making your life difficult (I think) is the `Option`, because it's what making you introduce an additional `async` block there. You could perhaps change `ShutdownSignal` to _not_ be an `Option`, and instead set it to [`std::future::pending`](https://doc.rust-lang.org/std/future/fn.pending.html) when `None`. That way your `select!` arm can simply be:

```rust
_ = &mut self.shutdown_signal => {}
```

and that arm will simply never resolve if no shutdown signal was given, and you don't need an `Option` _or_ an `Arc<Mutex`.

jonhoo (changes requested):
Hmm, it's a good point. I think I agree that `run_one` also shouldn't be possible once terminated :+1:

src/worker/mod.rs
rustworthy: Actually option never bothered me a lot, it is rather the _mutable self_ that we already need for the heartbeat task and  so we cannot we give it to any other arm. You can image now why I was suggesting to take the signal initially (but I really missed that one scenario where the worker will be re-used), and why I am suggesting to put it behind a tokio mutex now. I think it is fine in this case where we _know_ there is only one arm wanting this lock and so there will be no performance issues here.

As for the `std::future::pending`, I believe it is an elegant construct, but I am not sure the shutdown signal arm should be polled at all if they've never provided the signal. I am not sure that the runtime can inspect this future and see that it will _never_ resolve and so it should not waste _any_ efforts on it. Plus we are spawning workers (`Worker::for_worker`) and will need to pass `Box::pin(std::future::pending())` for those "processing" workers that actually do not need the signalling future (only the coordinator does).

jonhoo: Ah, and that's because Rust doesn't realize that `self.listen_for_heartbeats` doesn't actually need to mutable borrow all the _fields_ of `self`. Which is what the `Option` was trying to get around. Okay, I'm with you now. In that case, I think we should rather bring back the `Option` than have an unnecessary `Arc<Mutex`. It's not so much about the performance as it is about the cognitive complexity — we now need to remember why it is that thing has to be reference counted _and_ locked, and we now introduce the potential for deadlocks if we later start also taking it somewhere else.

A permanently-pending arm should indeed never cause any waste — that future will never have its waker awoken, so it will never be polled again. Wouldn't worry on that part about using `std::future::pending`.

rustworthy: yeah I wish there were a "sharded" mutable reference and I guess the compiler will support it the future.

As for the option and the mutex, I am not sure those are related. The option actually never helped in this situation (reminder: my taking a signal from an option and not putting it back was a _mistake_), it is the counted async mutex that did. As for the deadlocks - as long as `Worker::run` takes a _mutable_ reference - we are ok. I cannot imagine how they/we manage to cause a dead lock.

wdyt?

jonhoo: I figured it was easier to show than try to explain; here's what I meant: https://github.com/jonhoo/faktory-rs/pull/57/commits/8de6397a497e6bab43ae522ea1490582c032b289 (pushed directly to this branch).

If you're happy with it, I think we're good to merge!

rustworthy: Oh boy I could have simply put it back :facepalm:

rustworthy: honestly I never considered it since I was sure I've moved and there is no way I can get it in the other arm

rustworthy: oh.. and _now_ I see how the async block in your earlier comment is related :)

jonhoo: One more tweak: https://github.com/jonhoo/faktory-rs/pull/57/commits/6eb5b1726af3ac0a20861a29f51a0fb75f79e5b5

And yeah, that's why I need the `&mut shutdown_signal` in the `select!` arm — so that the arm doesn't _move_ `shutdown_signal`, it only borrows it. Once you _enter_ an arm in a `select!`, all the other arms are cancelled, and no longer borrow whatever they were borrowing.

jonhoo: Also note that while there is an `Option` there now, it's used somewhat differently than in your original. It's no longer "none == no shutdown signaler"; `None` is simply used as a temporary holding value.

jonhoo (approved):

Issue #56, opened 2024-03-01
Removing/clearing the queue programmatically


rbozan:
It is not clear to me how to remove the queue in Rust using the library. Basically, the equivalent of the Clear button in the WebUI :

![image](https://github.com/jonhoo/faktory-rs/assets/7997154/d4807200-1916-438f-b33f-1412bea73c0a)

jonhoo:
cc @rustworthy in case you know

rustworthy:
@rbozan thanks for pointing at that. Indeed, not all queue control actions are supported right now. We support pause/resume queues by names, and plan to cover the rest of functionality:
- pause/resume all queues;
- clear (effectively, _remove_) all queues;
- clear (effectively, _remove_) queues by names;

Planning to deliver that after the ongoing work on the switch to async is finished. But a PR will be much appreciated!
---
For reference:

The relevant Faktory server [code](https://github.com/contribsys/faktory/blob/main/server/commands.go#L45-L75)
The corresponding [section](https://github.com/jonhoo/faktory-rs/blob/main/src/proto/single/cmd.rs#L263-L295) in our bindings.

---
cc @jonhoo

jonhoo:
Closed by https://github.com/jonhoo/faktory-rs/pull/59

PR #59, opened 2024-04-09
Support remove queues, as well as pause all / resume all

rustworthy:
Addresses #56

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/59)


tests/real/community.rs
```diff
@@ -139,38 +139,189 @@ async fn fail() {
...
+
+    // though there _was_ a job in each queue, consuming from
+    // the removed queues will not yield anything
+    assert!(!worker.run_one(0, &[local_1]).await.unwrap());
+    assert!(!worker.run_one(0, &[local_2]).await.unwrap());
+    assert!(!rx.try_recv().is_ok());
+
+    // let's inspect the sever state again
+    let server_state = client.info().await.unwrap();
+    let queues = server_state.get("faktory").unwrap().get("queues").unwrap();
```
rustworthy, line 240: this can be more ergonomic, please see #60

src/proto/client/mod.rs
```diff
@@ -359,10 +373,10 @@ where
         Ok((jobs_count - errors.len(), Some(errors)))
     }
 
-    /// Retrieve information about the running server.
+    /// Retrieve [information](crate::ServerState) about the running server.
     ///
     /// The returned value is the result of running the `INFO` command on the server.
-    pub async fn info(&mut self) -> Result<serde_json::Value, Error> {
+    pub async fn info(&mut self) -> Result<single::ServerState, Error> {
```
rustworthy, line 379: This is breaking and if we decide it's worth it  - these changes should join the upcoming breaking 'async' release.
We expose an _additional_ method like `get_server_state` which would be returning `ServerState`, but this would be confusing for the lib user.

src/proto/single/resp.rs
```diff
@@ -117,6 +120,88 @@ pub async fn read_ok<R: AsyncBufReadExt + Unpin>(r: R) -> Result<(), Error> {
...
+    /// A mapping between a queue name and its size (number of jobs on the queue).
+    /// The keys of this map effectively make up a list of queues that are currently
+    /// registered in the Faktory service.
+    pub queues: HashMap<String, u64>,
+
+    /// Faktory's task runner stats.
+    ///
+    /// Note that this is exposed as a "generic" `serde_json::Value` since this info
+    /// belongs to deep implementation details of the Faktory service.
+    pub tasks: serde_json::Value,
```
rustworthy, line 151: This is kept as `serde_json::Value` since those are super deep implementation details:
```sh
tasks: Object {
            "Busy": Object {
                "reaped": Number(0),
                "size": Number(0),
            },
            "Dead": Object {
                "cycles": Number(0),
                "enqueued": Number(0),
                "size": Number(0),
                "wall_time_sec": Number(0),
            },
            "Retries": Object {
                "cycles": Number(1),
                "enqueued": Number(0),
                "size": Number(0),
                "wall_time_sec": Number(0.000170441),
            },
            "Scheduled": Object {
                "cycles": Number(1),
                "enqueued": Number(0),
                "size": Number(0),
                "wall_time_sec": Number(0.000316549),
            },
            "Workers": Object {
                "reaped": Number(0),
                "size": Number(1),
            },
        },
```

If they need those - they definitely will find their way in via `serde_json::Values::get`.

rustworthy: wdyt?

tests/real/community.rs
```diff
@@ -61,6 +61,73 @@ async fn roundtrip() {
...
+    // because if you have not pruned its volume the Faktory will still keep the queue name
+    // as registered.
+    // But generally, we are performing a clean-up by consuming the jobs from the local queue/
+    // and then deleting the queue programmatically, so there is normally no need to prune docker
+    // volumes to perform the next test run. Also note that on CI we are always starting a-fresh.
+    let server_state = client.info().await.unwrap();
+    assert_eq!(*server_state.faktory.queues.get(local).unwrap(), 0);
+    assert!(server_state.faktory.total_processed >= 1); // at least 1 job from this test
+
+    client.queue_remove(&[local]).await.unwrap();
```
rustworthy, line 119: Now that we got this command, this can be utilized for clean ups

While not crucial on CI, can make life easier when developing locally: there will be no remainders from the previous test runs.

jonhoo:
going to hold off on reviewing this until after #49

src/proto/single/resp.rs
rustworthy: The `tasks` field is [commented as deprecated](https://github.com/contribsys/faktory/pull/474/files#diff-4bbbde2a77215c352f9e99f0f2764fb9fb09eaa9052433b58a474b33019f7486R23), and so I guess not only should this stay as generic `serde_json::Value` but also annotated as `#[deprecated]`

rustworthy:
@jonhoo 

I encountered some issues when trying to `make Client hold a boxed dyn Connection` . Our `Reconnect` trait mentions `Self` in the return value of it's only method, making the whole thing not object safe. So introducing a `Client` which is
a struct with the following shape ...
```rust
type Connection = Box<FaktoryConnection>;
struct Client {
  connection: Connection,
  opts: ClientOptions,
 }
 ```
... did not work. 

I might be missing something though..

Otherwise the PR is ready for review.

jonhoo:
Would you mind splitting this PR into multiple, one for each logical change? It's a little hard to review the changes when multiple of them are bundled together in this way.

rustworthy:
> Would you mind splitting this PR into multiple, one for each logical change? It's a little hard to review the changes when multiple of them are bundled together in this way.

Sure

rustworthy:
@jonhoo 

Only queue control actions are left in this PR. Please have a look at it once again

jonhoo (approved):
This looks great, no notes!

Issue #60, opened 2024-04-10
Make info command return ServerState struct


rustworthy:
Info command will now return a result with `serde_json::Value` which made a lot of sense during the period when the `Faktory` library was being actively developed and so the shape of the return type could change breakingly. The last time when that [interface](https://github.com/contribsys/faktory/blob/main/server/server.go#L433-L452) was updated was a couple of years [ago](https://github.com/contribsys/faktory/commit/4a57c90a844b6cd448cbc51cc14c7462e0de3539) and so we may want to introduce a dedicated type for it to make `Client::info` more ergonomic:

```rust
let server_state = client.info().await.unwrap(); // ServerState
let queues = &server_state.faktory.queues; // &HashMap<String, u64>
```

While currently - provided they know the keys - this will look like:
```rust
let server_state = client.info().await.unwrap(); // serde_json::Value
let queues = server_state.get("faktory").unwrap().get("queues").unwrap(); // &serde_json::Value
```

An example usage of `Client::info` can be found [here](https://github.com/jonhoo/faktory-rs/pull/59/files#diff-a07cf223951f9e59e9ab517b9c9b35ee99fc360ab0616a13aac62ffa2347143bR224-R225)

Since a breaking change is occurring due to switch to async, the `Client::info` can be updated to return `ServerState`.

jonhoo:
That seems reasonable! If we also annotate it with `#[non_exhaustive]`, we'll even be able to expose more fields if they should be added.

mperham:
Related, see my recent Faktory PR which adds a typed structure for INFO.

jonhoo:
Ah, this is https://github.com/contribsys/faktory/pull/474?

Given we're doing a breaking change as part of #49 anyway, I think we should probably just remove `.info` entirely and add `.current_info` that returns a type matching the one you added here:
https://github.com/contribsys/faktory/pull/474/files#diff-4bbbde2a77215c352f9e99f0f2764fb9fb09eaa9052433b58a474b33019f7486R9

rustworthy:
> Ah, this is [contribsys/faktory#474](https://github.com/contribsys/faktory/pull/474)?
> 
[...]

PR #63, opened 2024-05-20
Add ServerState and Client::current_info

rustworthy:
Addresses #60

We will also add sets to DataSnapshot struct as soon as the Faktory server [starts returning it](https://github.com/contribsys/faktory/pull/474/files#diff-78f42ba40d0f10b08c73b7e6bb8376f398e249c963cf549e89591d6b6826b9a4R451-R456).

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/63)

### DIFF
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -153,7 +153,7 @@ pub struct DataSnapshot {
     pub tasks: serde_json::Value,
 }
 
-/// Faktory's server process stats.
+/// Faktory's server process information.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct ServerSnapshot {
     /// Description of the server process (e.g. "Faktory").
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -201,8 +201,15 @@ pub struct FaktoryState {
     /// Server time.
     pub now: DateTime<Utc>,
 
-    /// Server time as a string formatted as "%H:%M:%S UTC" (e.g. "19:47:39 UTC").
-    pub server_utc_time: String,
+    /// Server time (naive representation).
+    ///
+    /// Faktory sends it as a string formatted as "%H:%M:%S UTC" (e.g. "19:47:39 UTC")
+    /// and it is being parsed as `NaiveTime`.
+    ///
+    /// Most of the time, though, you will want to use [`FaktoryState::now`] instead.
+    #[serde(deserialize_with = "utils::deser_server_time")]
+    #[serde(serialize_with = "utils::ser_server_time")]
+    pub server_utc_time: chrono::naive::NaiveTime,
 
     /// Faktory service information.
     #[serde(rename = "faktory")]
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -149,7 +149,9 @@ pub struct DataSnapshot {
     ///
     /// Note that this is exposed as a "generic" `serde_json::Value` since this info
     /// belongs to deep implementation details of the Faktory service.
-    #[deprecated]
+    #[deprecated(
+        note = "marked as deprecated in the Faktory source code and is likely to be completely removed in the future, so please do not rely on this data"
+    )]
     pub tasks: serde_json::Value,
 }
 
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -125,6 +125,10 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
 // ----------------------------------------------
 
 /// Faktory service information.
+///
+/// This holds information on the registered [queues](DataSnapshot::queues) as well as
+/// some aggregated data, e.g. total number of jobs [processed](DataSnapshot::total_processed),
+/// total number of jobs [enqueued](DataSnapshot::total_enqueued), etc.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[non_exhaustive]
 pub struct DataSnapshot {
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -156,7 +156,7 @@ pub struct DataSnapshot {
 /// Faktory's server process stats.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct ServerSnapshot {
-    /// Faktory's description (e.g. "Faktory").
+    /// Description of the server process (e.g. "Faktory").
     pub description: String,
 
     /// Faktory's version as semver (e.g. "1.8.0").
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -1,7 +1,7 @@
 use super::utils;
 use crate::error::{self, Error};
 use chrono::{DateTime, Utc};
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::time::Duration;
 use tokio::io::{AsyncBufRead, AsyncBufReadExt, AsyncReadExt};
 
@@ -149,7 +149,7 @@ pub struct DataSnapshot {
     /// A mapping between a queue name and its size (number of jobs on the queue).
     /// The keys of this map effectively make up a list of queues that are currently
     /// registered in the Faktory service.
-    pub queues: HashMap<String, u64>,
+    pub queues: BTreeMap<String, u64>,
 
     /// ***Deprecated***. Faktory's task runner stats.
     ///
--- a/src/proto/single/resp.rs
+++ b/src/proto/single/resp.rs
@@ -1,8 +1,10 @@
+use crate::error::{self, Error};
+use chrono::{DateTime, Utc};
+use std::collections::HashMap;
+use tokio::io::{AsyncBufRead, AsyncBufReadExt, AsyncReadExt};
+
 #[cfg(feature = "ent")]
 use crate::ent::BatchId;
-
-use crate::error::{self, Error};
-use tokio::io::AsyncBufRead;
 
 pub fn bad(expected: &'static str, got: &RawResponse) -> error::Protocol {
     let stringy = match *got {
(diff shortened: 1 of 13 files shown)


src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+    /// A mapping between a queue name and its size (number of jobs on the queue).
+    /// The keys of this map effectively make up a list of queues that are currently
+    /// registered in the Faktory service.
+    pub queues: HashMap<String, u64>,
+
+    /// ***Deprecated***. Faktory's task runner stats.
+    ///
+    /// Note that this is exposed as a "generic" `serde_json::Value` since this info
+    /// belongs to deep implementation details of the Faktory service.
+    #[deprecated]
```
jonhoo, line 152: Let's include a message for this deprecation as well.

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+
+    /// Total number of queues.
+    pub total_queues: u64,
+
+    /// Queues stats.
+    ///
+    /// A mapping between a queue name and its size (number of jobs on the queue).
+    /// The keys of this map effectively make up a list of queues that are currently
+    /// registered in the Faktory service.
+    pub queues: HashMap<String, u64>,
```
jonhoo, line 146: Do you think iterating over these in order would ever be interesting to callers? As in, is there an argument for making this a `BTreeMap` instead?

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+    /// Note that this is exposed as a "generic" `serde_json::Value` since this info
+    /// belongs to deep implementation details of the Faktory service.
+    #[deprecated]
+    pub tasks: serde_json::Value,
+}
+
+/// Faktory's server process stats.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ServerSnapshot {
+    /// Faktory's description (e.g. "Faktory").
```
jonhoo, line 159:
```suggestion
    /// Description of the server process (e.g. "Faktory").
```

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+    pub description: String,
+
+    /// Faktory's version as semver (e.g. "1.8.0").
```
jonhoo, line 162: If this is actually semver, should we use [`semver::Version`](https://docs.rs/semver/latest/semver/struct.Version.html) as the type here instead?

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+    #[serde(rename = "faktory_version")]
+    pub version: String,
+
+    /// Faktory server process uptime in seconds.
+    pub uptime: u64,
```
jonhoo, line 167: Can we expose this as `std::time::Duration` instead so units aren't going to confuse people (i.e., by using `#[serde(deserialize_with = )]` to construct it from seconds)?

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
     Err(bad("server ok", &rr).into())
 }
 
+// ----------------------------------------------
+
+/// Faktory service stats.
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
jonhoo, line 126: Any particular reason to also implement `Serialize` here, or can we remove that?

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+
+    /// ***Deprecated***. Faktory's task runner stats.
+    ///
+    /// Note that this is exposed as a "generic" `serde_json::Value` since this info
+    /// belongs to deep implementation details of the Faktory service.
+    #[deprecated]
+    pub tasks: serde_json::Value,
+}
+
+/// Faktory's server process stats.
```
jonhoo, line 156:
```suggestion
/// Faktory's server process information.
```

src/proto/single/resp.rs
jonhoo: When will this ever not be "Faktory"?

src/proto/single/resp.rs
jonhoo: Since it's not just stats

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+/// let _server_state = client.current_info().await.unwrap();
+/// # });
+/// ```
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct FaktoryState {
+    /// Server time.
+    pub now: DateTime<Utc>,
+
+    /// Server time as a string formatted as "%H:%M:%S UTC" (e.g. "19:47:39 UTC").
+    pub server_utc_time: String,
```
jonhoo, line 199: Can/should we convert this into a [`NaiveTime`](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveTime.html)?

Why is it useful for this when we already have `now` above?

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
     Err(bad("server ok", &rr).into())
 }
 
+// ----------------------------------------------
+
+/// Faktory service stats.
```
jonhoo, line 125: Can we be more descriptive here to distinguish this from `ServerSnapshot`? I _think_ this specifically is statistics related to the job queues (as opposed to the server process)?

src/proto/single/resp.rs
```diff
@@ -118,6 +120,92 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+
+    /// Server time as a string formatted as "%H:%M:%S UTC" (e.g. "19:47:39 UTC").
+    pub server_utc_time: String,
+
+    /// Faktory service stats.
+    #[serde(rename = "faktory")]
+    pub data: DataSnapshot,
+
+    /// Faktory's server process stats.
+    pub server: ServerSnapshot,
```
jonhoo, line 206: Let's make the comments here mirror what we end up making the docstring for each of the types.

tests/real/community.rs
```diff
@@ -70,6 +70,74 @@ async fn roundtrip() {
...
+    let mut client = Client::connect(None).await.unwrap();
+
+    // examine server state before pushing anything
+    let server_state = client.current_info().await.unwrap();
+    assert!(server_state.data.queues.get(local).is_none());
+    // the following two assertions are not super-helpful but
+    // there is not much info we can make meaningful assetions on anyhow
+    // (like memusage, server description string, version, etc.)
+    assert!(server_state.server.connections >= 2); // at least two clients from the current test
+    assert!(server_state.server.uptime > 0); // if IPC is happenning, this should hold :)
```
jonhoo, line 96:
```suggestion
    assert_ne!(server_state.server.uptime, 0); // if IPC is happenning, this should hold :)
```

tests/real/community.rs
```diff
@@ -70,6 +70,74 @@ async fn roundtrip() {
...
+    // prepare a producing client
+    let mut client = Client::connect(None).await.unwrap();
+
+    // examine server state before pushing anything
+    let server_state = client.current_info().await.unwrap();
+    assert!(server_state.data.queues.get(local).is_none());
+    // the following two assertions are not super-helpful but
+    // there is not much info we can make meaningful assetions on anyhow
+    // (like memusage, server description string, version, etc.)
+    assert!(server_state.server.connections >= 2); // at least two clients from the current test
```
jonhoo, line 95:
```suggestion
    assert!(server_state.server.connections >= 2, "{}", server_state.server.connections); // at least two clients from the current test
```
so that we get to see what the count was if this fails

tests/real/community.rs
```diff
@@ -70,6 +70,74 @@ async fn roundtrip() {
...
+        )
+        .await
+        .unwrap();
+
+    // we only pushed 1 job on this queue
+    let server_state = client.current_info().await.unwrap();
+    assert_eq!(*server_state.data.queues.get(local).unwrap(), 1);
+    assert!(server_state.data.total_enqueued >= 1); // at least 1 job from this test
+    assert!(server_state.data.total_queues >= 1); // at least 1 qeueu from this test
+                                                  // let's know consume that job ...
```
jonhoo, line 114: I suppose we could check here that the number is at least +1 compared to what it was when we read it just before enqueueing the job?

Also: typo "queue"

jonhoo (changes requested):
Generally looks good, mostly nits!

src/proto/single/resp.rs
rustworthy: I can easily imagine it be `Faktory Enterprise` or `Faktory Valkey`.

src/proto/single/resp.rs
rustworthy: As a user, I actually would expect most of the things to be `Serialize` for me to be able to send them further

src/proto/single/resp.rs
rustworthy: Done!

src/proto/single/resp.rs
rustworthy: I feel like the `Client::current_info` is more for monitoring tools, as so I actually can imagine a situation where they are fetching this data on the server within the framework supporting SSR (not necessarily this, but for the example's sake) and iterating over these in a, say, `view!` macro. Normally, upon submitting a task via UI and having been redirected to the list view page, people will expect the queues to preserve the same order. We can of course apply some sorting onto the `HashMap` entries at an earlier step , but we get it (at a low price) right away with `BTreeMap`. Not sure how contrived this example is though.

Another example would be  an alert tool (maybe a headless one), that iterates over the `DataSnapshot::queues` and creates an event in the number of jobs on an queue is more than the predefined alert threshold. In this case, though, the order is of no particular interest.

Wdyt?

src/proto/single/resp.rs
rustworthy: EDIT: even though it is now simply ["Faktory"](https://github.com/rustworthy/faktory/blob/ec73667a31f31242fbbbab9dc5d2fd7cf74c3ce9/client/faktory.go#L4)

src/proto/single/resp.rs
rustworthy: Done!

src/proto/single/resp.rs
rustworthy: Done. Also bebug output would be
```
server: ServerSnapshot {
        description: "Faktory",
        version: Version {
            major: 1,
            minor: 8,
            patch: 0,
        },
        uptime: 6s,
        connections: 2,
        command_count: 1,
        used_memory_mb: 7,
    },
```
This `6s` fmt is just killing :smiley:

src/proto/single/resp.rs
rustworthy: Updated. As for this duplication, looks like `server_utc_time` is maintained for backward compatibility. 

Added to the docs, encouraging people to use `now`.

src/proto/single/resp.rs
rustworthy: wdyt?

tests/real/community.rs
rustworthy: also added a couple of helpers: `assert_gt`, `assert_lt`, etc.

src/proto/single/resp.rs
rustworthy: Exactly, queues and some aggregated data. Updated.

src/proto/single/resp.rs
jonhoo: So, the main challenge is that because this field is `pub`, we can't change its type later. As such, I'd be inclined to make it a "maximally useful" type, even if it comes with a slight runtime overhead. So let's switch to `BTreeMap` :+1:

src/proto/single/resp.rs
jonhoo: I suppose for forwarding, or maybe also for mocking. Okay, fair enough! :+1:

tests/real/community.rs
```diff
@@ -88,12 +88,24 @@ async fn server_state() {
 
     // examine server state before pushing anything
     let server_state = client.current_info().await.unwrap();
+    // the Faktory release we are writing bindings and testing
+    // against is at least "1.8.0"
+    assert_gte!(server_state.server.version.major, 1);
```
jonhoo, line 93: I would probably make this `assert_eq!`, since if we get a bump to `2`, the `>=` on `minor` would no longer work right.

src/proto/single/resp.rs
```diff
@@ -118,6 +122,107 @@ pub async fn read_ok<R: AsyncBufRead + Unpin>(r: R) -> Result<(), Error> {
...
+
+    /// Total number of queues.
+    pub total_queues: u64,
+
+    /// Queues stats.
+    ///
+    /// A mapping between a queue name and its size (number of jobs on the queue).
+    /// The keys of this map effectively make up a list of queues that are currently
+    /// registered in the Faktory service.
+    pub queues: HashMap<String, u64>,
```
jonhoo, line 152:
```suggestion
    pub queues: BTreeMap<String, u64>,
```

src/proto/single/resp.rs
```diff
@@ -1,9 +1,13 @@
+use super::utils;
+use crate::error::{self, Error};
+use chrono::{DateTime, Utc};
+use std::collections::HashMap;
```
jonhoo, line 4:
```suggestion
use std::collections::BTreeMap;
```

jonhoo (changes requested):

tests/real/community.rs
jonhoo, line 93:
```suggestion
    assert_eq!(server_state.server.version.major, 1);
```

jonhoo (approved):

PR #64, opened 2024-05-20
Tech debt

rustworthy:
Threads that needed to be addressed after the recent "async" PR merge:

- [x] only require references from the bindings' user where we could do without owning values internally as per this [discussion](https://reviewable.io/reviews/jonhoo/faktory-rs/49#-NvzaTNB5u0502mF7RVA);
- [x] make Client hold a boxed dyn Connection as per [thread](https://reviewable.io/reviews/jonhoo/faktory-rs/49#-NuJNBtf8ftQKWeYPRbD);
- [x] a tech debt from the recent merge as per [thread](https://reviewable.io/reviews/jonhoo/faktory-rs/49#-NxRty-qDzpHT2XpIeEc);
- [x] address https://github.com/jonhoo/faktory-rs/pull/65#discussion_r1677116648

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/64)


rustworthy:
As for the boxed dyn Connection thread, please see this [comment](https://github.com/jonhoo/faktory-rs/pull/59#issuecomment-2109362878)

jonhoo:
For the `Box<dyn Connection>` bit, I _think_ the trick is going to be to make `Reconnect` also return a `Box<dyn Connection>`. That should make it object safe!

rustworthy:
> For the `Box<dyn Connection>` bit, I _think_ the trick is going to be to make `Reconnect` also return a `Box<dyn Connection>`. That should make it object safe!

That actually did the trick. We've been able to get rid of the `S` generic, please have a look at this.

src/tls/rustls.rs
```diff
@@ -121,13 +124,32 @@ where
...
+    async fn reconnect(&mut self) -> io::Result<proto::BoxedConnection> {
+        let stream = self.get_mut().stream.get_mut().0.reconnect().await?;
+        let tls_stream = TlsStream::new(
+            stream,
+            self.get_ref().connector.clone(),
+            self.get_ref().hostname.clone(),
+        )
+        .await?;
+        let buffered = BufStream::new(tls_stream);
+        Ok(Box::new(buffered))
```
rustworthy, line 152: feels like there is some redundancy in here..

src/worker/builder.rs
```diff
@@ -152,15 +152,19 @@ impl<E: 'static> WorkerBuilder<E> {
...
-    pub async fn connect_with<S: AsyncRead + AsyncWrite + Send + Unpin>(
+    pub async fn connect_with<S>(
         mut self,
         stream: S,
         pwd: Option<String>,
-    ) -> Result<Worker<BufStream<S>, E>, Error> {
+    ) -> Result<Worker<E>, Error>
+    where
+        S: AsyncRead + AsyncWrite + Send + Sync + Unpin + 'static,
+        BufStream<S>: Reconnect,
```
jonhoo, line 162: I _think_ if we have a blanket implementation of `Reconnect` for `BufStream<S> where S: Reconnect` it should be sufficient to add `Reconnect` to the bounds on `S`

src/worker/builder.rs
```diff
@@ -152,15 +152,19 @@ impl<E: 'static> WorkerBuilder<E> {
...
+    {
         self.opts.password = pwd;
         self.opts.is_worker = true;
         let buffered = BufStream::new(stream);
```
jonhoo, line 166: How do you feel about _not_ wrapping the provided stream in `BufStream` here and instead encouraging the caller to do so? My thinking is that they may _already_ have some stream type that does buffering, and we then force them here to have _two_ layers of buffering.

src/tls/rustls.rs
```diff
@@ -121,13 +124,32 @@ where
...
 
 #[async_trait::async_trait]
-impl<S> Reconnect for TlsStream<S>
-where
-    S: AsyncRead + AsyncWrite + Send + Unpin + Reconnect,
-{
-    async fn reconnect(&mut self) -> io::Result<Self> {
-        let stream = self.stream.get_mut().0.reconnect().await?;
-        TlsStream::new(stream, self.connector.clone(), self.hostname.clone()).await
+impl Reconnect for BufStream<TlsStream<tokio::net::TcpStream>> {
```
jonhoo, line 127: Can we have a blanket:
```rust
impl<S> Reconnect for BufStream<S> where S: Reconnect
```

and then just `impl Reconnect for TlsStream<TcpStream>`?

jonhoo: In fact, maybe we should _also_ have:
```rust
impl<S> Reconnect for TlsStream<S> where S: Reconnect
```
rather than specializing this to `TlsStream<TcpStream>`?

src/tls/rustls.rs
jonhoo: Should be helped by the comments above I think

src/tls/rustls.rs
```diff
@@ -121,13 +124,32 @@ where
...
+            self.get_ref().hostname.clone(),
+        )
+        .await?;
+        let buffered = BufStream::new(tls_stream);
+        Ok(Box::new(buffered))
+    }
+}
+
+#[async_trait::async_trait]
+impl Reconnect for BufStream<TlsStream<proto::BoxedConnection>> {
```
jonhoo, line 142: This `impl` can probably go away with the two blanket impls I've outlined above (at least assuming we `impl Reconnect for BoxedConnection`).

src/tls/rustls.rs
jonhoo: This would mean for `native_tls` we also only need a single impl and that's

```rust
impl<S> Reconnect for TlsStream<S> where S: Reconnect
```

(with `native_tls::TlsStream`)

src/proto/single/ent/progress.rs
```diff
@@ -38,6 +38,12 @@ pub struct ProgressUpdate {
     pub reserve_until: Option<DateTime<Utc>>,
 }
 
+impl AsRef<Self> for ProgressUpdate {
```
jonhoo, line 41: Reading <https://doc.rust-lang.org/std/convert/trait.AsRef.html#relation-to-borrow>, I'm starting to wonder if we should prefer `Borrow` over `AsRef` for the functions that just want a reference. That would also (I believe) mean this `impl` could go away.

src/proto/single/ent/cmd.rs
```diff
@@ -3,27 +3,36 @@ use crate::error::Error;
...
-            }
-            Self::Get(jid) => {
-                w.write_all(b"TRACK GET ").await?;
-                w.write_all(jid.as_bytes()).await?;
-                Ok(w.write_all(b"\r\n").await?)
-            }
-        }
+        w.write_all(b"TRACK GET ").await?;
+        w.write_all(self.0.as_ref().as_bytes()).await?;
+        Ok(w.write_all(b"\r\n").await?)
```
jonhoo, line 36: Nice!

src/worker/builder.rs
jonhoo: Ditto for `Client::connect_with`

src/proto/mod.rs
```diff
@@ -28,16 +33,17 @@ pub use batch::{Batch, BatchBuilder, BatchHandle, BatchStatus, CallbackState};
 
 /// A stream that can be re-established after failing.
 #[async_trait::async_trait]
-pub trait Reconnect: Sized {
+pub trait Reconnect {
```
jonhoo, line 36: I think you'll want to add a blanket impl of this trait for `Box`:
```rust
impl<S> Reconnect for Box<S> where S: Reconnect
```

src/proto/client/ent.rs
```diff
@@ -1,20 +1,25 @@
...
+impl Client {
     /// Send information on a job's execution progress to Faktory.
-    pub async fn set_progress(&mut self, upd: ProgressUpdate) -> Result<(), Error> {
-        let cmd = Track::Set(upd);
-        self.issue(&cmd).await?.read_ok().await
+    pub async fn set_progress<P>(&mut self, upd: P) -> Result<(), Error>
+    where
+        P: AsRef<ProgressUpdate> + Sync,
+    {
+        self.issue(&upd).await?.read_ok().await
```
jonhoo, line 14: If you change this to
```suggestion
        self.issue(upd.as_ref()).await?.read_ok().await
```
I think you can get rid of the
```rust
impl<P> FaktoryCommand for P
where
    P: AsRef<ProgressUpdate> + Sync,
```
in favor of just a `impl FaktoryCommand for ProgressUpdate`

jonhoo (changes requested):
I'm loving this set of changes!

src/worker/builder.rs
rustworthy: the thing with `Reconnect` is that `Box<dyn Connection> returned from `async fn reconnect()` should already be `AsyncBufRead`. So we are accepting any `S` that is `AsyncRead + AsyncWrite + Send + Sync + Unpin + 'static'` because we then buffer it making it `AsyncBufRead` which only then qualifies it as `Reconnect`

src/worker/builder.rs
rustworthy: My thinking was that requiring `AsyncBufRead` is not something standard/predictable, this is why we only ask them for AsyncRead/AsyncWrite. But I also agree that if they already have got a buffered one - maybe it makes sense to offer them an optional way. Looking into it

rustworthy: Added `connect_with_buffered` for both `Client` and `WorkerBuilder` with the according docs. Also updated one of the tests to use this method just as a sanity check. The ordinary `connect_with` will be  performing the buffering internally and passing the stuff through to `connect_with_buffered`, wdyt?

src/proto/mod.rs
rustworthy: done

src/proto/client/ent.rs
rustworthy: Updated

src/tls/rustls.rs
rustworthy: please see [comment](https://github.com/jonhoo/faktory-rs/pull/64#discussion_r1702745387)

src/proto/single/ent/progress.rs
rustworthy: Well, I this would go away indeed and being Borrow will do the job internally. I guess the question is, how ergonomic and expected this is. Here is how the `get_progress_ would look like:
```rust
pub async fn get_progress<J>(&mut self, jid: J) -> Result<Option<Progress>, Error>
    where
        J: Borrow<JobId> + Sync,
    {
        let cmd = FetchProgress::new(jid.borrow());
        self.issue(&cmd).await?.read_json().await
    }
```

Say I've got a struct keeping JobId and and TraceId, for which the AsRef<JobId> is implemented. It would be nice if I could pass it to `Client::get_progress` which is not possible due to:
```rust
```

rustworthy:
@jonhoo  Sorry for thrashing the loop :sweat_smile: 

Please have another look at this PR.  I think I've addressed almost all the threads, and will need your advice in a couple of them.

---

Something I've noticed, but would like to discuss with you first:

- Job's [`reserve_for`](https://github.com/jonhoo/faktory-rs/blob/a63b658c338bf95c0ab19ba62d2260715e803149/src/proto/single/mod.rs#L113) can be `Duration`

- What do you make of [handler](https://github.com/jonhoo/faktory-rs/blob/a63b658c338bf95c0ab19ba62d2260715e803149/src/worker/runner.rs#L55) accepting something which is `Into<Job>`? This way we could for example supply a macro that will implement `Into<Job>` for something that is `Serialize` .  Just thinking about the future  :)

- `Worker` should probably have a [`Worker::is_terminated`](https://github.com/jonhoo/faktory-rs/blob/a63b658c338bf95c0ab19ba62d2260715e803149/src/worker/mod.rs#L164)  method being publicly available, so that they can check if they can [safely](https://github.com/jonhoo/faktory-rs/blob/a63b658c338bf95c0ab19ba62d2260715e803149/src/worker/mod.rs#L418) re-run it.

- We could use tokio's [interval](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.tick) in the heartbeat thread when we are checking the state [every 100 millis](https://github.com/jonhoo/faktory-rs/blob/a63b658c338bf95c0ab19ba62d2260715e803149/src/worker/health.rs#L41). That's a nit, but _maybe_ a more idiomatic variant.

src/worker/builder.rs
jonhoo: I think I actually prefer just having a single `connect_with` and noting in its docs that you should probably wrap the passed-in stream in `BufStream::new` if it doesn't already implement `AsyncBufRead`. After all, most folks will only use this method if they specifically have a custom stream already!

src/proto/single/ent/progress.rs
jonhoo: Oh, right, `Borrow` is overloaded to also mean equivalence for `Hash` and friends. That's awkward. Okay, then I'm fine with sticking with `AsRef` :+1:

tests/real/community.rs
```diff
@@ -729,22 +729,38 @@ async fn test_panic_in_handler() {
...
     let mut w = Worker::builder::<io::Error>()
-        .register_blocking_fn("panic", |_j| {
+        .register_blocking_fn("panic_SYNC_handler", |_j| {
             panic!("Panic inside the handler...");
         })
+        .register_fn("panic_ASYNC_handler", |j| async move {
+            // potentially going out of bounds
+            let arg1 = &j.args()[0];
+            let arg2 = &j.args()[1];
+            let _ = arg1.as_i64().unwrap() + arg2.as_i64().unwrap();
```
jonhoo, line 739: Can we instead have this take a single `bool` arg that is whether or not it should call `panic!`?

src/worker/mod.rs
```diff
@@ -225,11 +228,23 @@ impl<E: StdError + 'static + Send> Worker<E> {
             .get(job.kind())
             .ok_or(Failed::BadJobType(job.kind().to_string()))?;
         match handler {
-            Callback::Async(cb) => cb.run(job).await.map_err(Failed::Application),
+            Callback::Async(cb) => AssertUnwindSafe(cb.run(job))
```
jonhoo, line 231: Do we know that this is unwind safe? I feel like this should perhaps be a bound on what `cb.run` returns instead? Alternatively, can we `tokio::spawn` here, which will catch the panic for us?

Cargo.toml
```diff
@@ -50,6 +50,7 @@ tokio-rustls = { version = "0.25.0", optional = true }
 tracing = "0.1"
 url = "2"
 semver = { version = "1.0.23", features = ["serde"] }
+futures = "0.3.30"
```
jonhoo, line 53: The `futures` crate is, to my knowledge, mostly unmaintained these days, so I'd like to avoid taking a dependency on it.

tests/real/community.rs
```diff
@@ -729,22 +729,38 @@ async fn test_panic_in_handler() {
...
+        })
         .connect(None)
         .await
         .unwrap();
 
-    Client::connect(None)
-        .await
-        .unwrap()
-        .enqueue(Job::builder("panic").queue(local).build())
+    let mut c = Client::connect(None).await.unwrap();
```
jonhoo, line 746: The more I look at `connect(None)`, the less I like it. I wonder if we should have two methods:
```rust
fn connect() { .. }
fn connect_to(addr: &str) { .. }
```
what do you think?

tests/real/community.rs
rustworthy: I actually _wanted_ to do it this way: we are testing here how we downcast the panic message. This is as opposed to our calling `panic!` explicitly (which we are also doing in this test)

src/worker/builder.rs
jonhoo: Ah, hmm, I think we might need two traits here then:
```rust
trait Reconnect {
    fn reconnect(&mut self) -> io::Result<BoxedConnection>;
}
trait UnbufferedConnection: AsyncRead + AsyncWrite + Send + Sync + Unpin + 'static + ReconnectUnbuffered {}
impl<T> UnbufferedConnection for T where T: AsyncRead + AsyncWrite + Send + Sync + Unpin + 'static {}
trait ReconnectUnbuffered {
    fn reconnect(&mut self) -> io::Result<Box<dyn UnbufferedConnection>>;
}
```
with
```rust
impl<S> Reconnect for BufStream<S> where S: UnbufferedConnection {}
```
it's not super pretty, but it would allow folks to bring their own buffered _or_ unbuffered connection types, and use the former directly or use the latter wrapped in a `BufStream`.

tests/real/community.rs
jonhoo: I think you can still test that yourself by using [`panic_any`](https://doc.rust-lang.org/std/panic/fn.panic_any.html), which allows passing in an arbitrary type to raise as the panic (thus testing downcasting).

src/worker/builder.rs
rustworthy: updated

src/worker/mod.rs
rustworthy: We do not and yes this should be a bound if we want to be safe here. I like your suggestion with `tokio::spawn` better. I remember this [comment](https://github.com/jonhoo/faktory-rs/pull/49#discussion_r1468867348) and so I think now _may_ want to spawn just to immediately await, we just should be unwrap the result, rather handle it properly.

The spawning here also brings me now back to this [thread](https://github.com/jonhoo/faktory-rs/pull/65#discussion_r1614435096) :smile:

Cargo.toml
rustworthy: And we are also not relying on any of its functionality any longer. Removed

tests/real/community.rs
rustworthy: I think the suggested changes will make the bindings more ergonomic, because now even though they are already providing FAKTORY_URL into the environment, we still make them pass `None` to the the `Client::connect`.

What do you make of batching these changes with my earlier [suggestions](https://github.com/jonhoo/faktory-rs/pull/64#issuecomment-2278094341) in a follow-up "polishing" PR? Also what do you make of those suggestions?

src/worker/mod.rs
rustworthy: indeed I ended up using your earlier suggestion to avoid double-arc'ing

tests/real/community.rs
rustworthy: No that we are relaying this to `tokio::spawn`, this test restored to its previous state plus panicking inside the async handler.

src/worker/builder.rs
rustworthy: I am afraid I am not following here.

For both `Client::connect_with` and `WorkerBuilder::connect_with` the bound is now:
```
S: AsyncBufRead + AsyncWrite + Reconnect + Send + Sync + Unpin + 'static,
```

tests/real/community.rs
jonhoo: Yeah, I'm fine doing that change in a separate PR — this one is already growing quite large and unwieldy! I should have used reviewable from the start for it :p

src/worker/builder.rs
jonhoo: I think maybe I've led you astray. I read over the code again this morning, and I think I understand what you've been trying to tell me :sweat_smile: I think the only thing left that's confusing to me is how we can have
```rust
impl<S> Reconnect for BufStream<S>
```
and also
```rust
impl Reconnect for BufStream<TlsStream<tokio::net::TcpStream>>
```
and
```rust
impl Reconnect for BufStream<TlsStream<proto::BoxedConnection>>
```
It feels like that first `impl` should overlap with the other two. I'm guessing it doesn't because we define `Reconnect` ourselves and we don't have
```rust
impl Reconnect for TlsStream<tokio::net::TcpStream>
```
or
```rust
impl Reconnect for TlsStream<proto::BoxedConnection>
```
That makes me wonder though — would it be sufficient to instead have
```rust
impl Reconnect for TlsStream<S>
``
(where I don't think we even need to require `AsyncBufRead` for that `S`)?

It also feels weird that we even have
```rust
impl Reconnect for BufStream<TlsStream<proto::BoxedConnection>>
```
Because that suggests we may end up double-buffering (since `BoxedConnection` is already `AsyncBufRead`).

jonhoo (approved):
I'm going to approve and land this just so we can hone the discussion (in a separate PR) specifically onto the one remaining question around the `Reconnect` impls. I still think we need to nail it down before an eventual release.

PR #65, opened 2024-05-20
Add support for blocking handlers

rustworthy:
Rationale:

We probably should not be discriminating those who have been using the crate for CPU intensive tasks running on the background, e.g. image processing.

We are not adding "fully-fledged" `SyncJobRunner` trait, just to keep things simple. `WorkerBuilder::register_blocking_fn` should be considered an escape hatch. Adding `SyncJobRunner` later on - should this be required  - will not be a breaking change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/65)

### DIFF
--- a/src/worker/builder.rs
+++ b/src/worker/builder.rs
@@ -115,9 +115,12 @@ impl<E: 'static> WorkerBuilder<E> {
     /// [docs](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) for
     /// how to set the upper limit on the number of threads in the mentioned pool and other details.
     ///
-    /// Note that it is not recommended to mix blocking and non-blocking tasks and so if many of your
-    /// handlers are blocking, you will want to launch a dedicated worker process (at least a separate
-    /// Tokio Runtime) where only blocking handlers will be used.
+    /// You can mix and match async and blocking handlers in a single `Worker`. However, note that
+    /// there is no active management of the blocking tasks in `tokio`, and so if you end up with more
+    /// CPU-intensive blocking handlers executing at the same time than you have cores, the asynchronous
+    /// handler tasks (and indeed, all tasks) will suffer as a result. If you have a lot of blocking tasks,
+    /// consider using the standard async job handler (which you can register with [`WorkerBuilder::register`]
+    /// or [`WorkerBuilder::register_fn`]) and add explicit code to manage the blocking tasks appropriately.
     ///
     /// Also note that only one single handler per job kind is supported. Registering another handler
     /// for the same job kind will silently override the handler registered previously.
--- a/src/worker/builder.rs
+++ b/src/worker/builder.rs
@@ -4,6 +4,7 @@ use crate::{
     Error, Job, JobRunner, WorkerId,
 };
 use std::future::Future;
+use std::sync::Arc;
 use tokio::io::{AsyncRead, AsyncWrite, BufStream};
 use tokio::net::TcpStream as TokioStream;
 
@@ -106,6 +107,30 @@ impl<E: 'static> WorkerBuilder<E> {
         self.register(kind, Closure(handler))
     }
 
+    /// Register a _blocking_ (synchronous) handler function for the given job type (`kind`).
+    ///
+    /// This is an analogue of [`register_fn`](WorkerBuilder::register_fn) for compute heavy tasks.
+    /// Internally, `tokio`'s `spawn_blocking` is used when a job arrives whose type matches `kind`,
+    /// and so the `handler` is executed in a dedicated pool for blocking operations. See `Tokio`'s
+    /// [docs](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) for
+    /// how to set the upper limit on the number of threads in the mentioned pool and other details.
+    ///
+    /// Note that it is not recommended to mix blocking and non-blocking tasks and so if many of your
+    /// handlers are blocking, you will want to launch a dedicated worker process (at least a separate
+    /// Tokio Runtime) where only blocking handlers will be used.
+    ///
+    /// Also note that only one single handler per job kind is supported. Registering another handler
+    /// for the same job kind will silently override the handler registered previously.
+    pub fn register_blocking_fn<K, H>(mut self, kind: K, handler: H) -> Self
+    where
+        K: Into<String>,
+        H: Fn(Job) -> Result<(), E> + Send + Sync + 'static,
+    {
+        self.callbacks
+            .insert(kind.into(), super::Callback::Sync(Arc::new(handler)));
+        self
+    }
+
     /// Register a handler for the given job type (`kind`).
     ///
     /// Whenever a job whose type matches `kind` is fetched from the Faktory, the given handler
@@ -118,7 +143,8 @@ impl<E: 'static> WorkerBuilder<E> {
         K: Into<String>,
         H: JobRunner<Error = E> + 'static,
     {
-        self.callbacks.insert(kind.into(), Box::new(runner));
+        self.callbacks
+            .insert(kind.into(), super::Callback::Async(Box::new(runner)));
         self
     }
 
--- a/src/worker/mod.rs
+++ b/src/worker/mod.rs
@@ -6,7 +6,7 @@ use std::sync::{atomic, Arc};
 use std::{error::Error as StdError, sync::atomic::AtomicUsize};
 use tokio::io::{AsyncBufRead, AsyncWrite};
 use tokio::net::TcpStream;
-use tokio::task::{spawn_blocking, AbortHandle, JoinSet};
+use tokio::task::{spawn_blocking, AbortHandle, JoinError, JoinSet};
 
 mod builder;
 mod health;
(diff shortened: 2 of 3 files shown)


src/worker/mod.rs
```diff
@@ -20,7 +20,12 @@ pub(crate) const STATUS_RUNNING: usize = 0;
 pub(crate) const STATUS_QUIET: usize = 1;
 pub(crate) const STATUS_TERMINATING: usize = 2;
 
-type CallbacksRegistry<E> = FnvHashMap<String, runner::BoxedJobRunner<E>>;
+pub(crate) enum Callback<E> {
+    Async(runner::BoxedJobRunner<E>),
+    Sync(Arc<dyn Fn(Job) -> Result<(), E> + Sync + Send + 'static>),
```
jonhoo, line 25: I think this can be `Box<dyn` since the callbacks are already stored in an `Arc`

src/worker/builder.rs
```diff
@@ -106,6 +107,30 @@ impl<E: 'static> WorkerBuilder<E> {
...
+    ///
+    /// This is an analogue of [`register_fn`](WorkerBuilder::register_fn) for compute heavy tasks.
+    /// Internally, `tokio`'s `spawn_blocking` is used when a job arrives whose type matches `kind`,
+    /// and so the `handler` is executed in a dedicated pool for blocking operations. See `Tokio`'s
+    /// [docs](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) for
+    /// how to set the upper limit on the number of threads in the mentioned pool and other details.
+    ///
+    /// Note that it is not recommended to mix blocking and non-blocking tasks and so if many of your
+    /// handlers are blocking, you will want to launch a dedicated worker process (at least a separate
+    /// Tokio Runtime) where only blocking handlers will be used.
```
jonhoo, line 120: I'm not sure I expect this to be necessary. As long as we're using `spawn_blocking`, it should be fine to have some async and some sync handlers.

src/worker/mod.rs
```diff
@@ -187,7 +192,16 @@ impl<S: AsyncBufRead + AsyncWrite + Send + Unpin, E: StdError + 'static + Send>
             .callbacks
             .get(job.kind())
             .ok_or(Failed::BadJobType(job.kind().to_string()))?;
-        handler.run(job).await.map_err(Failed::Application)
+        match handler {
+            Callback::Async(cb) => cb.run(job).await.map_err(Failed::Application),
+            Callback::Sync(cb) => {
+                let cb = Arc::clone(cb);
```
jonhoo, line 198: Ah, this is why you `Arc` each individual `Sync` function. You could instead just `Arc::clone` all of `handler` I think to avoid the extra layer of `Arc`.

src/worker/mod.rs
```diff
@@ -187,7 +192,16 @@ impl<S: AsyncBufRead + AsyncWrite + Send + Unpin, E: StdError + 'static + Send>
...
+                spawn_blocking(move || cb(job))
+                    .await
+                    .expect("joined ok")
```
jonhoo, line 201: This will fail if the provided `Fn` ever panics. May be worth bubbling up as another kind of error (`Failed`) instead?

tests/real/community.rs
```diff
@@ -340,3 +341,47 @@ async fn test_jobs_created_with_builder() {
...
+        .register_blocking_fn("cpu_intensive", |_j| {
+            // Imagine some compute heavy operations:serializing, sorting, matrix multiplication, etc.
+            std::thread::sleep(Duration::from_millis(1000));
+            Ok::<(), io::Error>(())
+        })
+        .register_fn("io_intensive", |_j| async move {
+            // Imagine fetching data for this user from various origins,
+            // updating an entry on them in the database, and then sending them
+            // an email and pushing a follow-up task on the Faktory queue
+            Ok::<(), io::Error>(())
```
jonhoo, line 363: Maybe stick a `tokio::sleep` in here?

jonhoo (changes requested):
Mostly smaller bits!

src/worker/mod.rs
rustworthy: Hmm, you mean having the sync handler boxed as advised [here](https://github.com/jonhoo/faktory-rs/pull/65#discussion_r1614433393) and  getting the `self.callbacks` (?) cloned instead? Because the individual handlers are not arced in the registry, while the registry itself is arced on the `Worker`. 

I might be missing something, please have a look:

```rust
async fn run_job(&mut self, job: Job) -> Result<(), Failed<E>> {
        let handler = self
            .callbacks
            .get(job.kind())
            .ok_or(Failed::BadJobType(job.kind().to_string()))?;
        match handler {
            Callback::Async(cb) => cb.run(job).await.map_err(Failed::Application),
            _ => {
                let callbacks = Arc::clone(&self.callbacks);
                spawn_blocking(move || {
                    if let Callback::Sync(cb) = callbacks
                        .get(job.kind())
                        .expect("sync handler to be available")
                    {
                        cb(job)
                    } else {
                        unreachable!();
                    }
                })
                .await
                .expect("joined ok")
                .map_err(Failed::Application)
            }
        }
    }
```
We are indeed avoiding introducing new arcs in this case, instead we cloning the whole of the registry. It just looks weird a little bit, wdyt?

src/worker/mod.rs
rustworthy: Please see [comment](https://github.com/jonhoo/faktory-rs/pull/65#discussion_r1615257651)

src/worker/builder.rs
rustworthy: This is my reading of this [section](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code) and namely this passage:
```
If your code is CPU-bound and you wish to limit the number of threads used to run it, you should use a separate thread pool dedicated to CPU bound tasks. For example, you could consider using the [rayon](https://docs.rs/rayon) library for CPU-bound tasks. It is also possible to create an extra Tokio runtime dedicated to CPU-bound tasks, but if you do this, you should be careful that the extra runtime runs only CPU-bound tasks, as IO-bound tasks on that runtime will behave poorly.
```
So "if many handlers are blocking" they _can_  employ one single runtime and mix blocking handlers with async ones, since we are using `spawn_blocking` (and so are protecting the cooperative scheduling for async tasks), but we are encouraging them to _consider_ launching a dedicated runtime for their blocking tasks. Or you think this note is redundant?

tests/real/community.rs
rustworthy: Indeed, that will also be illustrative: blocking vs non-blocking sleep.

src/worker/mod.rs
```diff
@@ -176,18 +181,28 @@ impl<S: AsyncWrite + Send + Unpin, E> Worker<S, E> {
...
 impl<S: AsyncBufRead + AsyncWrite + Send + Unpin, E: StdError + 'static + Send> Worker<S, E> {
-    async fn run_job(&mut self, job: Job) -> Result<(), Failed<E>> {
+    async fn run_job(&mut self, job: Job) -> Result<(), Failed<E, JoinError>> {
         let handler = self
             .callbacks
             .get(job.kind())
             .ok_or(Failed::BadJobType(job.kind().to_string()))?;
-        handler.run(job).await.map_err(Failed::Application)
+        match handler {
+            Callback::Async(cb) => cb.run(job).await.map_err(Failed::Application),
```
rustworthy, line 197: I am worried that this is not the case for async handlers, panicking in there is not intercepted in the way is now being done for `register_blocking_fn`. I wonder if might need to use `futures::future::FutureExt::catch_unwind` for those futures, not sure..

src/worker/mod.rs
rustworthy: added a local error variant

src/worker/builder.rs
jonhoo: I think giving a nudge in this direction is fine, although I think the nudge is currently too strong. I wonder if we just want to say something like:

> You can mix and match async and blocking handlers in a single `Worker`. However, note that there is no active management of the blocking tasks in tokio, and so if you end up with more CPU-intensive blocking handlers executing at the same time than you have cores, the asynchronous handler tasks (and indeed, all tasks) will suffer as a result. If you have a lot of blocking tasks, consider using the standard async job handler and add explicit code to manage the blocking tasks appropriately.

src/worker/mod.rs
jonhoo: Yeah, that's exactly what I was thinking. I agree it looks pretty weird, and it's sad that we have to double-match. I think we should just stick with what we had with the double-`Arc`, even though it makes me sad.

src/worker/mod.rs
jonhoo: Let's do that in a follow-up (maybe the tech debt one?)

jonhoo (approved):
Looks good subject to the one documentation change

src/worker/mod.rs
rustworthy: pinned in there

src/worker/builder.rs
rustworthy: updated

PR #66, opened 2024-05-20
Add tracing and examples for `run` and `run_to_completion`

rustworthy:
This  PR is also adding:
- tracing dep and emitting heartbeat event at trace level - so that we know that we are healthy;
- examples `run` and `run_to_completion`;

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/66)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,20 @@ if let Err(e) = w.run(&["default"]).await {
 }
 ```
 
-## Run test suite locally
+Also see some usage examples in `examples` directory in the project's root. You can run an example with:
 
+```bash
+cargo run --example example_name
+```
+
+For instance, to run a `run_to_completion` example in release mode, hit:
+
+```bash
+cargo run --example run_to_completion --release
+```
+Make sure you've got Faktory server up-and-running. See [instructions](#run-test-suite-locally) on how to spin up Faktory locally.
+
+## Run test suite locally
 First ensure the "Factory" service is running and accepting connections on your machine.
 To launch it a [Factory](https://hub.docker.com/r/contribsys/faktory/) container with [docker](https://docs.docker.com/engine/install/), run:
 
--- a/examples/run.rs
+++ b/examples/run.rs
@@ -10,10 +10,8 @@ pub struct JobHandler<T> {
 }
 
 impl<T> JobHandler<T> {
-    pub fn new(chan: &Arc<mpsc::Sender<T>>) -> Self {
-        Self {
-            chan: Arc::clone(chan),
-        }
+    pub fn new(chan: Arc<mpsc::Sender<T>>) -> Self {
+        Self { chan }
     }
 }
 
--- a/examples/run.rs
+++ b/examples/run.rs
@@ -68,7 +68,7 @@ async fn main() {
 
     // create a worker
     let mut w = Worker::builder()
-        .register("job_type", JobHandler::new(&tx))
+        .register("job_type", JobHandler::new(Arc::clone(&tx)))
         .connect(None)
         .await
         .expect("Connected to server");
--- a/examples/run.rs
+++ b/examples/run.rs
@@ -18,17 +18,9 @@ impl<T> JobHandler<T> {
 impl JobHandler<u64> {
     async fn process_one(&self, job: Job) -> IOResult<()> {
         time::sleep(time::Duration::from_millis(100)).await;
-        let mut args = job.args().into_iter();
-        let x = args
-            .next()
-            .expect("'x' to be some")
-            .as_u64()
-            .expect("'x' to be an integer");
-        let y = args
-            .next()
-            .expect("'y' to be some")
-            .as_u64()
-            .expect("'y' to be an interger");
+        let args = job.args();
+        let x = args[0].as_u64().expect("'x' to be an integer");
+        let y = args[1].as_u64().expect("'y' to be an interger");
         self.chan.send(x + y).await.expect("no error");
         Ok(())
     }
@@ -79,7 +71,7 @@ async fn main() {
     let res = rx.recv().await.expect("some calculation result");
 
     tracing::info!(
-        "Send a job with args `vec![5, 8]`. Receiced result `{}`",
+        "Send a job with args `vec![5, 8]`. Received result `{}`",
         res
     );
 }
--- a/examples/run.rs
+++ b/examples/run.rs
@@ -70,7 +70,7 @@ async fn main() {
     // wait for processing results
     let res = rx.recv().await.expect("some calculation result");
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -62,8 +62,20 @@ if let Err(e) = w.run(&["default"]).await {
...
+```bash
+cargo run --example example_name
+```
+
+For instance, to run a `run_to_completion` example in release mode, hit:
 
+```bash
+cargo run --example run_to_completion --release
+```
+Make sure you've got Faktory server up-and-running. See [instructions](#run-test-suite-locally) on how to spin up Faktory locally.
```
jonhoo, line 76: Great instructions, thank you!

examples/run.rs
```diff
@@ -0,0 +1,87 @@
...
+pub struct JobHandler<T> {
+    chan: Arc<mpsc::Sender<T>>,
+}
+
+impl<T> JobHandler<T> {
+    pub fn new(chan: &Arc<mpsc::Sender<T>>) -> Self {
+        Self {
+            chan: Arc::clone(chan),
+        }
+    }
```
jonhoo, line 17:
```suggestion
    pub fn new(chan: Arc<mpsc::Sender<T>>) -> Self {
        Self { chan }
    }
```
nit: the convention in Rust is to allow the caller to be in charge of getting an owned instance of a value

examples/run.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+impl JobHandler<u64> {
+    async fn process_one(&self, job: Job) -> IOResult<()> {
+        time::sleep(time::Duration::from_millis(100)).await;
+        let mut args = job.args().into_iter();
```
jonhoo, line 23: Why go via an iterator instead of just directly indexing into `job.args()`?

examples/run.rs
```diff
@@ -0,0 +1,87 @@
...
+    async fn run(&self, job: Job) -> Result<(), Self::Error> {
+        self.process_one(job).await.unwrap();
+        Ok(())
+    }
+}
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt::fmt()
+        .with_max_level(tracing::Level::TRACE)
```
jonhoo, line 52: Why this?

examples/run.rs
```diff
@@ -0,0 +1,87 @@
...
+        .expect("client successfully connected");
+
+    // push one job onto the default queue
+    c.enqueue(Job::builder("job_type").args(vec![5, 8]).build())
+        .await
+        .expect("enqueued ok");
+
+    // create a worker
+    let mut w = Worker::builder()
+        .register("job_type", JobHandler::new(&tx))
```
jonhoo, line 73:
```suggestion
        .register("job_type", JobHandler::new(Arc::clone(&tx)))
```

examples/run.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let _handle = tokio::spawn(async move { w.run(&["default"]).await });
+
+    // wait for processing results
+    let res = rx.recv().await.expect("some calculation result");
+
+    tracing::info!(
+        "Send a job with args `vec![5, 8]`. Receiced result `{}`",
+        res
+    );
```
jonhoo, line 86: This feels like it could just be a `println!` since we're in an example.

src/worker/health.rs
```diff
@@ -8,6 +8,8 @@ use std::{
 use tokio::io::{AsyncBufRead, AsyncWrite};
 use tokio::time::sleep as tokio_sleep;
 
+const CHECK_STATE_INTERVAL_MILLIS: u64 = 100;
+const HEARTBEAT_INTERVAL_SECS: u64 = 5;
```
jonhoo, line 12: The `Duration` constructors (like `from_millis`) are `const`, so you can call them directly here and remove the unit from the const name :)
```suggestion
const CHECK_STATE_INTERVAL: Duration = Duration::from_millis(100);
const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
```

src/worker/health.rs
```diff
@@ -22,6 +24,10 @@ where
     ///   but should _continue_ processing its current job (if any);
     ///
     /// See more details [here](https://github.com/contribsys/faktory/blob/b4a93227a3323ab4b1365b0c37c2fac4f9588cc8/server/workers.go#L13-L49).
+    ///
+    /// Note that this method is not cancellation safe. We are using an interval timer internally, that
```
jonhoo, line 28: This is a good doc addition. Makes me wonder: do we call this method in the context of a `select!` ever at the moment? Or call it from an `async fn` that a user may `select!` over (and that we thus need to also document isn't cancellation safe)?

jonhoo (changes requested):
Mostly small nits!

examples/run.rs
rustworthy: Indeed, just like with the `&str` vs `String` case we came across elsewhere.

examples/run.rs
rustworthy: I guess I just applied the same pattern as with CLI args, they look identical. This encourages us to handle the `None` case and helps avoid out of range indexing. But for this example's sake, we can simplify things of course and index directly into the `job.args()`.

rustworthy: Updated

examples/run.rs
rustworthy: I really wanted to have [traces](https://github.com/jonhoo/faktory-rs/blob/a09d7c92abe2a0477c5193573fba06fba8a4eac1/src/worker/health.rs#L66) appear in the console.

examples/run.rs
rustworthy: Updated

src/worker/health.rs
rustworthy: Not just yet - this comment belongs to #57 where we do call in within a `select!`. Moved it there.

 As for the `Worker::run` (and `Worker::run_to_completion`) those are actually also not cancellation safe because of the `AsyncWriteExt::write_all`. It is also not safe to run the `Worker::run` in a loop due to the terminated worker scenario: if they try to re-run a terminated worker, we will [panic](https://github.com/jonhoo/faktory-rs/blob/3c966bdca641f4fc101d88b0098f6c99efbf7210/src/worker/mod.rs#L352-L355) right away.

rustworthy:
@jonhoo Please have a look at this PR once again. 

The tests are actually failing for the `main` branch, which is will be fixed in a dedicated PR: we simply need to adjust a couple of tests to use `Client::current_info` instead of the retired `Client::info`.

jonhoo (approved):
Looks good now — time to land 🛬

PR #67, opened 2024-05-24
Add 'precommit' task to Makefile

rustworthy:
That's a nit, but I've found myself repeating those commands over and over again.

I am not sure we need such stuff as [`pre-commit`](https://pre-commit.com/#rust) or [`husky`](https://github.com/typicode/husky). Just a utility command will do I guess.

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/67)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,13 @@ FAKTORY_PORT=7419
 FAKTORY_PORT_SECURE=17419
 FAKTORY_PORT_UI=7420
 
+.PHONT: precommit
+precommit: fmt check test/doc test/e2e test/e2e/tls
+
+.PHONY: fmt
+fmt:
+	cargo fmt
+
 .PHONY: check
 check:
 	cargo fmt --check
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ FAKTORY_PORT=7419
 FAKTORY_PORT_SECURE=17419
 FAKTORY_PORT_UI=7420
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,6 +4,13 @@ FAKTORY_PORT=7419
 FAKTORY_PORT_SECURE=17419
 FAKTORY_PORT_UI=7420
 
+.PHONT: precommit
+precommit: fmt check test/doc test/e2e test/e2e/tls
+
+.PHONY: fmt
+fmt:
+	cargo fmt
```
rustworthy, line 12: @jonhoo I remember you did not like this entry when we were first adding the make file, saying that we all have our formatting set up. I hope though that this command - when issued - will not interfere with any local set up, wdyt?

Makefile
```diff
@@ -4,6 +4,13 @@ FAKTORY_PORT=7419
 FAKTORY_PORT_SECURE=17419
 FAKTORY_PORT_UI=7420
 
+.PHONT: precommit
```
jonhoo, line 7:
```suggestion
.PHONY: precommit
```

Makefile
jonhoo: It's fine to have it here I suppose, it's more that I would expect people to generally have format-on-write turned on for Rust in their editors.

jonhoo (approved):

PR #71, opened 2024-06-22
Flaky test fix

rustworthy:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/71)

### DIFF
--- a/tests/real/community.rs
+++ b/tests/real/community.rs
@@ -105,9 +105,6 @@ async fn server_state() {
     ); // at least two clients from the current test
     assert_ne!(server_state.server.uptime.as_secs(), 0); // if IPC is happenning, this should hold :)
 
-    let nenqueued = server_state.data.total_enqueued;
-    let nqueues = server_state.data.total_queues;
-
     // push 1 job
     client
         .enqueue(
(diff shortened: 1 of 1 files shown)


tests/real/community.rs
```diff
@@ -125,21 +122,32 @@ async fn server_state() {
...
-    // `total_enqueued` should be at least +1 job from from last read
+
+    // It is tempting to make an assertion like `total enqueued this time >= total enqueued last time + 1`,
+    // but since we've got a server shared among numerous tests that are running in parallel, this
+    // assertion will not always work (though it will be true in the majority of test runs). Imagine
+    // a situation where between our last asking for server data state and now they have consumed all
+    // the pending jobs in _other_ queues. This is highly unlikely, but _is_ possible. So the only
+    // more or less guaranteed thing is that there is one pending job in our _local_ queue. It is "more or less"
+    // because another test _may_ still push onto and consume from this queue if we copypasta this test's
+    // contents and forget to update the local queue name, i.e. do not guarantee isolation at the queues level.
```
jonhoo, line 133: I think we should instead work pretty hard to ensure that tests indeed have different queue names, otherwise we're likely to run into much larger and hard-to-debug issues down the line. Happy to merge this as-is to resolve the flakiness, but if we can make the assertions stronger by upholding the invariant that every test has its own unique queue name, that's something I'd like us to do.

jonhoo (approved):

tests/real/community.rs
rustworthy: That's true and I guess we are already doing so using local queue names. In this case though the `server_state.data.total_enqueued` is the total number of currently enqueued jobs in the system rather than in a specific queue.

jonhoo: Ah, true — may be worth calling that out explicitly here

PR #73, opened 2024-07-10
Modify Worker::connect to use the password in the URL, as Client::connect does.

constlhq:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/73)


rustworthy (commented):
This is a great catch and the correct update, thank you!

This blunder has only been possible because I never added a test for the `WorkerBuilder` _authenticated_ flow. I've just added #75 as a primer and as soon  as a more ergonomic access to the `tls` feature will be enabled (which is now work-in-progress), this case will also be covered.

rustworthy:
I guess we need to fmt the updated module

rustworthy:
Hey @constlhq! Thanks again for this fix!

Will you bandwidth to iterate over this once again? 
The `fmt` job on the CI is failing, should be a quick fix.

jonhoo (approved):
Thanks!

PR #74, opened 2024-07-13
Add tls option to worker builder

rustworthy:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/74)


src/tls/native_tls.rs
```diff
@@ -62,6 +62,24 @@ impl TlsStream<TokioTcpStream> {
         .await
     }
 
+    /// Create a new TLS connection over TCP ***dangerously*** skipping TLS verification.
```
jonhoo, line 65:
```suggestion
    /// Create a new TLS connection over TCP **dangerously** skipping TLS verification.
```

src/tls/rustls.rs
```diff
@@ -66,6 +73,39 @@ impl TlsStream<TokioTcpStream> {
         TlsStream::with_connector(con, url).await
     }
 
+    /// Create a new TLS connection over TCP using native certificates.
+    ///
+    /// Unlike [`TlsStream::connect`], creates a root certificates store populated with the certificates
+    /// loaded from a platform-native certificate store. Thos method also allows to ***dangerously***
```
jonhoo, line 79:
```suggestion
    /// loaded from a platform-native certificate store. Thos method also allows to **dangerously**
```

src/tls/rustls.rs
```diff
@@ -66,6 +73,39 @@ impl TlsStream<TokioTcpStream> {
...
+    /// skip server certificates verification.
+    pub async fn connect_with_native_certs(
```
jonhoo, line 81: What's the rationale for us providing this rather than just providing a `connect_with_tls_config` that takes a `ClientConfig` directly which the user can configure however they would like? Or alternatively just telling users to use `TlsStream::new`/`TlsStream::with_connector` directly?

src/worker/builder.rs
```diff
@@ -18,6 +31,10 @@ pub struct WorkerBuilder<E> {
     callbacks: CallbacksRegistry<E>,
     shutdown_timeout: Option<Duration>,
     shutdown_signal: Option<ShutdownSignal>,
+    tls_kind: TlsKind,
+    #[cfg(any(feature = "native_tls", feature = "rustls"))]
+    #[cfg_attr(docsrs, doc(cfg(any(feature = "native_tls", feature = "rustls"))))]
+    verify_server_cert: bool,
```
jonhoo, line 37: I think I don't love us making `verify_server_cert` such a simple thing to turn off. I think I'd actually prefer that we make it very easy to connect with TLS normally (which you've already done with this PR), and then to say that any TLS customization requires going via `Worker::connect_with` and providing your own TLS connection.

jonhoo (changes requested):

src/worker/builder.rs
rustworthy: Hmm, I've used the [`imap`](https://github.com/jonhoo/rust-imap/blob/0f57af374669148a04a49f9151f83d4f5cb2ef34/src/client_builder.rs#L227-L229) crate for reference. I also think we are giving warnings that are standard, like `danger` prefixing and we are not encouraging people to turn it off.

Removing?

src/tls/rustls.rs
rustworthy: Well, I've saw they are doing so in the [`imap`](https://github.com/jonhoo/rust-imap/blob/8f3e678ee33caf39c6ec6135a2a79b192ca56620/src/client_builder.rs#L64-L72) crate, and I loved the idea of giving the user such a shortcut.

We ourselves making use of this `TlsStream::connect_with_native_certs` in the `WokrerBuilder` terminal method. Just handy, wdyt?

src/worker/builder.rs
jonhoo: I'm actually wondering whether it was a mistake for `imap` too :sweat_smile: For testing, one can simply run without TLS enabled (rather than using a self-signed cert). Hmm hmm hmm. I _think_ I'd prefer to leave this functionality out, at least for now, and then we can consider adding it if we get requests down the line.

rustworthy: dropped it and the related code

jonhoo (approved):
Nice, let's land it! :tada:

PR #75, opened 2024-07-13
Test password authentication in tls tests and workflow

rustworthy:
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/75)


rustworthy:
As soon as the #74 is finished and we start supporting specifying a desired tls option on the`WorkerBuilder` (rather than make them create a `TlsStream` and then provide that stream to the terminal `WorkerBuilder::connect_with`), we will "automatically" have the authenticated connection covered with tests for `Client` and `WorkerBuilder`.

jonhoo (approved):
Looks good to me, thanks!

PR #79, opened 2024-09-19
Tech debt

rustworthy:
- Use ::connect() vs ::connect_to(&str) as per [thread](https://github.com/jonhoo/faktory-rs/pull/64#discussion_r1720922173)
 
- Job's `reserve_for` is now using Duration;

- We are adding public  `Wokrer::is_terminated` so that they can check and _not_ run a terminated worker;

- Refactoring: using `tokio::time::interval` in the heartbeat thread instead of `tokio::time::sleep`

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/79)

### DIFF
--- a/src/proto/single/utils.rs
+++ b/src/proto/single/utils.rs
@@ -30,12 +30,34 @@ where
     serializer.serialize_u64(secs)
 }
 
+pub(crate) fn ser_optional_duration<S>(
+    value: &Option<Duration>,
+    serializer: S,
+) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    match value {
+        None => serializer.serialize_none(),
+        Some(dur) => serializer.serialize_u64(dur.as_secs()),
+    }
+}
+
 pub(crate) fn deser_duration<'de, D>(value: D) -> Result<Duration, D::Error>
 where
     D: Deserializer<'de>,
 {
     let secs = u64::deserialize(value)?;
     Ok(Duration::from_secs(secs))
+}
+
+pub(crate) fn deser_as_optional_duration<'de, D>(value: D) -> Result<Option<Duration>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(u64::deserialize(value)
+        .ok()
+        .map(|value| (Duration::from_secs(value))))
 }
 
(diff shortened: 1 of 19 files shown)


rustworthy:
@jonhoo Sorry it took me a while this time. I was moving apartment and preparing some legal docs around my expat life :sweat_smile: 

So I have addressed  - I _think_ so - all the threads, but [this one](https://github.com/jonhoo/faktory-rs/pull/64#discussion_r1739996721). I was guided by the compiler and the current implementation made it happy, and you unhappy :smile:  I agree it looks and feels weird, but I think I will need your help with this.

Please let me know if you while reviewing you catch other things that we better refactor now if this will be breaking.

src/proto/single/utils.rs
```diff
@@ -30,6 +30,19 @@ where
     serializer.serialize_u64(secs)
 }
 
+pub(crate) fn ser_optional_duration<S>(
```
jonhoo, line 33:
```suggestion
pub(crate) fn ser_optional_duration_in_seconds<S>(
```
the "in seconds" part is the main bit that requires this helper function. ditto for `deser`.

jonhoo (approved):
These changes all look great, thanks! A small nit is all.

No worries with the delay — I've had a similar delay now, and will likely be gone for the next month or so between work and holidays :sweat_smile:

jonhoo:
Feel free to merge this yourself once corrected :+1: 

I've also now given you publish rights to `faktory` on crates.io — you're now a full-fledged maintainer :)

jonhoo:
On the `Reconnect` thing, it's probably not ultimately all that important. My main wonder is if we can replace
```rust
impl Reconnect for BufStream<TlsStream<tokio::net::TcpStream>>
impl Reconnect for BufStream<TlsStream<proto::BoxedConnection>>
```
with
```rust
impl<S> Reconnect for TlsStream<S> where S: Reconnect
```

I'm also a little suspicious of
```rust
impl Reconnect for BufStream<TlsStream<proto::BoxedConnection>>
```
in the first place since that will end up returning a `BoxedConnection` that in turn _contains_ a `BoxedConnection`, meaning an extra (unnecessary) indirection.

Issue #61, opened 2024-04-29
Add support for Faktory MUTATE Api


rustworthy:
There are two ways to mutate server state: queue actions (pause, resume, clear) and via operations on the mutable sets (kill, discard, requeue, clear) jobs in a target set, where a target set is one of "scheduled", "dead", "retries"). We are currently supporting a subset of queue actions command and will support of them once #59 is landed.

See more details in the Faktory's wiki for [Mutate-API](https://github.com/contribsys/faktory/wiki/Mutate-API)  as well as
the Go bindings reference [implementation](https://github.com/contribsys/faktory/blob/b4a93227a3323ab4b1365b0c37c2fac4f9588cc8/client/mutate.go).

Also see how the sets stats will become available as part of the Faktory's current state structure once [this PR](https://github.com/contribsys/faktory/pull/474/files) is landed.

PR #82, opened 2024-11-02
[WIP] Bump deps

rustworthy:


Cargo.toml
```diff
@@ -50,8 +46,8 @@ tokio = { version = "1.35.1", features = [
     "time",            # anables `tokio::time` namespace holding the `sleep` utility and `Duraction` struct
 ] }
 tokio-native-tls = { version = "0.3.1", optional = true }
-tokio-rustls = { version = "0.25.0", optional = true }
-rustls-native-certs = { version = "0.7.1", optional = true }
+tokio-rustls = { version = "0.26", optional = true }
+rustls-platform-verifier = { version = "0.4", optional = true }
```
rustworthy, line 50: In their `0.8`, `rustls-native-certs` introduced changes that are [breaking](https://github.com/rustls/rustls-native-certs/commit/8aa350a8a0f02accd5be4652975a3fbc1b69bf97#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L120-R135) for us. 

We are now also taking into account the call from their [README](https://github.com/rustls/rustls-native-certs?tab=readme-ov-file)
"""
Instead of this crate, we suggest using [rustls-platform-verifier](https://github.com/rustls/rustls-platform-verifier), which provides a more robust solution with a simpler API. This crate is still maintained, but mostly for use inside the platform verifier on platforms where no other solution is available. 
"""

Makefile
```diff
@@ -54,17 +54,20 @@ test/doc:
...
+	cargo test --locked --all-features --all-targets -- \
+	--nocapture $(pattern)
 
 .PHONY: test/e2e/tls
 test/e2e/tls:
 	FAKTORY_URL_SECURE=tcp://:${FAKTORY_PASSWORD}@${FAKTORY_HOST}:${FAKTORY_PORT_SECURE} \
 	FAKTORY_URL=tcp://:${FAKTORY_PASSWORD}@${FAKTORY_HOST}:${FAKTORY_PORT} \
-	cargo test --locked --features native_tls,rustls --test tls -- --nocapture
+	cargo test --locked --features native_tls,rustls --test tls -- \
+	--nocapture $(pattern)
```
rustworthy, line 66: to be able to quickly run one specific test, very handy imho

Cargo.toml
```diff
@@ -16,24 +16,20 @@ exclude = [".github", "docker", ".gitignore", "Makefile"]
...
-]
+rustls = ["dep:pin-project", "dep:tokio-rustls", "dep:rustls-platform-verifier"]
 binaries = ["dep:clap", "tokio/macros"]
 ent = []
 
 [dependencies]
 async-trait = "0.1.77"
 clap = { version = "4.4.10", optional = true }
-chrono = { version = "0.4", features = [
+chrono = { version = "0.4.20", features = [
```
rustworthy, line 26: I think we were locked above this patch version, but still it's worth tightening due to potential [vulnerability](https://rustsec.org/advisories/RUSTSEC-2020-0159.html). Yay for `deps.rs` !

jonhoo: In general, vulnerabilities shouldn't be patched in consumers through their `Cargo.toml` bounds — the lockfile is for this. The manifest (`Cargo.toml`) is _just_ for declaring bounds on what's _possible_ to build with. We don't want to restrict this to a newer version unnecessarily as it would mean consumers that are locked on an older version for some reason are stuck without a path forward.

Cargo.toml
jonhoo, line 50: I've paged this out already (blame the holiday :sweat_smile:), but do we have types from either of these in our public API (esp. in trait implementations)? If so, this change will require a major version bump for faktory as well.

rustworthy: Our `TlsStream` [derefs](https://docs.rs/faktory/latest/faktory/rustls/struct.TlsStream.html#impl-Deref-for-TlsStream%3CS%3E) to `tokio_rustls`'  stream. 

Besides, [`TlsStream::with_client_config`](https://docs.rs/faktory/latest/faktory/rustls/struct.TlsStream.html#method.with_client_config) accepts a config struct from `rustls`. We are using `rustls` re-exported from `tokio_rustls`, and a major [bump](https://github.com/rustls/tokio-rustls/compare/rustls:5343443...rustls:82b4a3e#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R17) also took place there.

As for  `rustls-platform-verifier` - it's not part of out API.

rustworthy: So this probably - and unfortunately - should be a `0.14.0` release candidate.

Cargo.toml
rustworthy: Hmm I agree that the bounds should be kept as low as possible, especially if the library is not relying on  features added in minors and patches of its dependencies. However, a vulnerability  fix may (should?) be considered a feature we are relying upon, so I am not sure this is an "unnecessary" bump. It somehow feels not right.

And also can they really get locked on an older version in this case? If they could build their app with `0.4.0` till `0.4.19`, they _should_ be able to build with `0.4.20`.

Cargo.toml
rustworthy: Alternatively, we can  postpone these bumps and batch with other libs major versions updates that may come in the next couple of months.

I am now doing #61 and then will take #88. Bot of them are only extending what we currently got.

Cargo.toml
jonhoo: No, a vulnerability fix is not something we depend on :)

And yes, they can, such as if they use `<=` or `=` dependency specifiers.

Cargo.toml
jonhoo: Ah, that's unfortunate. Yeah, I think let's maybe hold off then in case we see other breaking changes coming down the line :+1:

Cargo.toml
rustworthy: this is technically possible, but I mean logically why would they go with `<=0.4.19` or `=0.4.19` for example :)

jonhoo: For example because the newer version requires a newer version of Rust than what they can currently run (eg, due to OS limitations or internal company policies).

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "faktory"
-version = "0.13.0"
+version = "0.13.1-rc0"
```
jonhoo, line 3: Let's make this `0.14.0-rc.1`

src/tls/rustls.rs
```diff
@@ -1,8 +1,10 @@
+use crate::error::Stream;
```
jonhoo, line 1: This looks like it ended up in the wrong place.

src/tls/rustls.rs
```diff
@@ -55,10 +57,12 @@ impl TlsStream<TokioTcpStream> {
     ///
     /// If `url` is given, but does not specify a port, it defaults to 7419.
     ///
-    /// Internally creates a `ClientConfig` with an _empty_ root certificates store and _no client
-    /// authentication_. Use [`with_client_config`](TlsStream::with_client_config)
-    /// or [`with_connector`](TlsStream::with_connector) for customized
-    /// `ClientConfig` and `TlsConnector` accordingly.
+    /// Internally, creates a [`ClientConfig`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html)
```
jonhoo, line 60:
```suggestion
    /// Internally, creates a [`ClientConfig`]
```

src/tls/rustls.rs
```diff
@@ -55,10 +57,12 @@ impl TlsStream<TokioTcpStream> {
...
+    /// with an _empty_ root certificates store and _no client authentication_.
+    ///
+    /// Use [`with_client_config`](TlsStream::with_client_config) or [`with_connector`](TlsStream::with_connector)
+    /// for customized `ClientConfig` and [`TlsConnector`](https://docs.rs/tokio-rustls/latest/tokio_rustls/struct.TlsConnector.html)
```
jonhoo, line 64:
```suggestion
    /// for customized `ClientConfig` and [`TlsConnector`]
```

src/tls/rustls.rs
```diff
@@ -120,10 +120,10 @@ where
 {
     /// Create a new TLS connection on an existing stream.
     ///
-    /// Internally creates a `ClientConfig` with an empty root certificates store and no client
-    /// authentication.
+    /// Internally, creates a [`ClientConfig`](https://docs.rs/tokio-native-tls/latest/tokio_native_tls/native_tls/struct.TlsConnector.html)
```
jonhoo, line 123:
```suggestion
    /// Internally, creates a [`ClientConfig`]
```

src/tls/rustls.rs
```diff
@@ -120,10 +120,10 @@ where
...
+    /// with an empty root certificates store and no client authentication.
     ///
-    /// Use [`new`](TlsStream::new) for a customized `TlsConnector`.
+    /// Use [`new`](TlsStream::new) for a customized [`TlsConnector`](https://docs.rs/tokio-rustls/latest/tokio_rustls/struct.TlsConnector.html).
```
jonhoo, line 126:
```suggestion
    /// Use [`new`](TlsStream::new) for a customized [`TlsConnector`].
```

PR #87, opened 2024-11-19
Mutate API

rustworthy:
addresses #61

this should also "unblock" #88 

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/87)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,7 +112,7 @@ pub use crate::proto::{
 
 /// Constructs used to mutate queues on the Faktory server.
 pub mod mutate {
-    pub use crate::proto::{Filter, FilterBuilder, JobSet};
+    pub use crate::proto::{Filter, JobSet};
 }
 
 pub use crate::worker::{JobRunner, StopDetails, StopReason, Worker, WorkerBuilder};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -106,8 +106,8 @@ mod worker;
 pub use crate::error::Error;
 
 pub use crate::proto::{
-    Client, Connection, DataSnapshot, FaktoryState, Job, JobBuilder, JobId, MutationFilter,
-    MutationFilterBuilder, MutationTarget, Reconnect, ServerSnapshot, WorkerId,
+    Client, Connection, DataSnapshot, Failure, FaktoryState, Job, JobBuilder, JobId,
+    MutationFilter, MutationFilterBuilder, MutationTarget, Reconnect, ServerSnapshot, WorkerId,
 };
 
 pub use crate::worker::{JobRunner, StopDetails, StopReason, Worker, WorkerBuilder};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -106,8 +106,8 @@ mod worker;
 pub use crate::error::Error;
 
 pub use crate::proto::{
-    Client, Connection, DataSnapshot, FaktoryState, Job, JobBuilder, JobId, Reconnect,
-    ServerSnapshot, WorkerId,
+    Client, Connection, DataSnapshot, FaktoryState, Job, JobBuilder, JobId, MutationFilter,
+    MutationFilterBuilder, MutationTarget, Reconnect, ServerSnapshot, WorkerId,
 };
 
 pub use crate::worker::{JobRunner, StopDetails, StopReason, Worker, WorkerBuilder};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,7 +112,7 @@ pub use crate::proto::{
 
 /// Constructs used to mutate queues on the Faktory server.
 pub mod mutate {
-    pub use crate::proto::{Filter, FilterBuilder, Target};
+    pub use crate::proto::{Filter, FilterBuilder, JobSet};
 }
 
 pub use crate::worker::{JobRunner, StopDetails, StopReason, Worker, WorkerBuilder};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -106,9 +106,14 @@ mod worker;
 pub use crate::error::Error;
 
 pub use crate::proto::{
-    Client, Connection, DataSnapshot, Failure, FaktoryState, Job, JobBuilder, JobId,
-    MutationFilter, MutationFilterBuilder, MutationTarget, Reconnect, ServerSnapshot, WorkerId,
+    Client, Connection, DataSnapshot, Failure, FaktoryState, Job, JobBuilder, JobId, Reconnect,
+    ServerSnapshot, WorkerId,
 };
+
+/// Constructs used to mutate queues on the Faktory server.
+pub mod mutate {
+    pub use crate::proto::{Filter, FilterBuilder, Target};
+}
 
 pub use crate::worker::{JobRunner, StopDetails, StopReason, Worker, WorkerBuilder};
 
--- a/src/proto/client/mutation.rs
+++ b/src/proto/client/mutation.rs
@@ -21,44 +21,100 @@ impl Client {
     /// client.requeue(MutationTarget::Retries, &filter).await.unwrap();
     /// # });
     /// ```
-    // For reference: https://github.com/contribsys/faktory/blob/10ccc2270dc2a1c95c3583f7c291a51b0292bb62/server/mutate.go#L35-L59
-    // The faktory will pull the entire targeted set from Redis to it's memory, iterate over
-    // each stringified job matching against "id":"...",  deserialize the matches into Jobs and
-    // re-queue those jobs.
     pub async fn requeue<'a, F>(&mut self, target: MutationTarget, filter: F) -> Result<(), Error>
     where
         F: Borrow<MutationFilter<'a>>,
     {
+        self.mutate(MutationType::Requeue, target, Some(filter.borrow()))
+            .await
+    }
+
+    /// Discard the jobs.
+    ///
+    /// Will throw the jobs away without any chance for re-scheduling
+    /// on the server side. If you want to still be able to process the jobs,
+    /// use [`Client::kill`] instead.
+    ///
+    /// E.g. to discard the currently enqueued jobs having "fizz" argument:
+    /// ```no_run
+    /// # tokio_test::block_on(async {
+    /// # use faktory::{Client, MutationTarget, MutationFilter};
+    /// # let mut client = Client::connect().await.unwrap();
+    /// let filter = MutationFilter::builder()
+    ///     .pattern(r#"*\"args\":\[\"fizz\"\]*"#)
+    ///     .build();
+    /// client.discard(MutationTarget::Scheduled, &filter).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn discard<'a, F>(&mut self, target: MutationTarget, filter: F) -> Result<(), Error>
+    where
+        F: Borrow<MutationFilter<'a>>,
+    {
+        self.mutate(MutationType::Discard, target, Some(filter.borrow()))
+            .await
+    }
+
+    /// Kil the jobs.
+    ///
+    /// Moves the jobs from the target structure to the `dead` set, meaning Faktory
+    /// will not touch it further unless you ask it to do so. You then can, for example,
+    /// manually process those jobs via the Web UI or send another mutation command
+    /// targeting [`MutationTarget::Dead`] set.
+    ///
+    /// E.g. to kill the currently enqueued jobs with "bill" argument:
+    /// ```no_run
+    /// # tokio_test::block_on(async {
+    /// # use faktory::{Client, MutationTarget, MutationFilter};
+    /// # let mut client = Client::connect().await.unwrap();
+    /// let filter = MutationFilter::builder()
+    ///     .pattern(r#"*\"args\":\[\"bill\"\]*"#)
+    ///     .build();
+    /// client.kill(MutationTarget::Scheduled, &filter).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn kill<'a, F>(&mut self, target: MutationTarget, filter: F) -> Result<(), Error>
+    where
+        F: Borrow<MutationFilter<'a>>,
+    {
+        self.mutate(MutationType::Kill, target, Some(filter.borrow()))
+            .await
+    }
+
+    /// Purge the targeted structure.
+    ///
+    /// Will have the same effect as [`Client::discard`] with an empty [`MutationFilter`],
+    /// but is special cased by Faktory and so is performed faster. Can be though of as
+    /// `TRUNCATE tablename` operation in the SQL world versus `DELETE FROM tablename`.
+    ///
+    /// E.g. to purged all the jobs that are pending in the `reties` set:
+    /// ```no_run
+    /// # tokio_test::block_on(async {
+    /// # use faktory::{Client, MutationTarget};
+    /// # let mut client = Client::connect().await.unwrap();
+    /// client.clear(MutationTarget::Retries).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn clear(&mut self, target: MutationTarget) -> Result<(), Error> {
+        self.mutate(MutationType::Clear, target, None).await
+    }
+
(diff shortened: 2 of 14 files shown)


src/lib.rs
```diff
@@ -106,8 +106,8 @@ mod worker;
 pub use crate::error::Error;
 
 pub use crate::proto::{
-    Client, Connection, DataSnapshot, FaktoryState, Job, JobBuilder, JobId, Reconnect,
-    ServerSnapshot, WorkerId,
+    Client, Connection, DataSnapshot, FaktoryState, Job, JobBuilder, JobId, MutationFilter,
+    MutationFilterBuilder, MutationTarget, Reconnect, ServerSnapshot, WorkerId,
 };
```
jonhoo, line 111: This is starting to be a lot of top-level exports. How do you feel about starting to group some of them the way we've done with the `pub mod ent` below? Like `pub mod mutate` for example? Then we can also rename them to avoid the `Mutation` name prefix.

src/proto/single/cmd.rs
```diff
@@ -42,12 +78,7 @@ where
 
 pub(crate) struct Info;
 
-#[async_trait::async_trait]
-impl FaktoryCommand for Info {
-    async fn issue<W: AsyncWrite + Unpin + Send>(&self, w: &mut W) -> Result<(), Error> {
-        Ok(w.write_all(b"INFO\r\n").await?)
-    }
-}
+self_to_cmd!(Info);
```
jonhoo, line 81: Nice, I like the macro expansion here.

src/proto/single/cmd.rs
```diff
@@ -311,3 +313,39 @@ impl<'a, S: AsRef<str>> QueueControl<'a, S> {
...
+        None => true,
+        Some(f) => f.is_empty(),
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+pub(crate) struct MutationAction<'a> {
+    pub(crate) cmd: MutationType,
+    pub(crate) target: MutationTarget,
+    #[serde(skip_serializing_if = "filter_is_empty")]
```
jonhoo, line 347: Is an empty-but-set mutation filter actually == to an unset filter? In some systems, the two have different semantics (usually, unset means "use the default").

src/proto/client/mutation.rs
```diff
@@ -0,0 +1,182 @@
...
+use std::borrow::Borrow;
+
+impl Client {
+    /// Re-enqueue the jobs.
+    ///
+    /// ***Warning!*** The `MUTATE` API is not supposed to be used as part of application logic,
+    /// you will want to use it for administration purposes only.
+    ///
+    /// This method will immediately move the jobs from the targeted set (see [`MutationTarget`])
+    /// to their queues. This will apply to the jobs satisfying the [`filter`](crate::MutationFilter).
```
jonhoo, line 14: I'm having a hard time following what this operation actually does. For example, with `MutationTarget::Scheduled`, what does it mean for the jobs to "immediately be moved to their queues"? They're already in queues?

src/proto/single/mutation.rs
```diff
@@ -0,0 +1,112 @@
...
+use derive_builder::Builder;
+
+#[cfg(doc)]
+use crate::{Client, Job};
+
+/// Mutation target set.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "lowercase")]
+#[non_exhaustive]
+pub enum MutationTarget {
```
jonhoo, line 11: Is this actually specific to mutations? Feels like a general purpose job set specifier to me.

It's also weird to me that the "mutation target" isn't just another filter, given that seems to be the function it has.

src/proto/client/mutation.rs
```diff
@@ -0,0 +1,182 @@
...
+    ///
+    /// ```no_run
+    /// # tokio_test::block_on(async {
+    /// # use faktory::{JobId, Client, MutationTarget, MutationFilter};
+    /// # let mut client = Client::connect().await.unwrap();
+    /// let job_id1 = JobId::new("3sgE_qwtqw1501");
+    /// let job_id2 = JobId::new("3sgE_qwtqw1502");
+    /// let failed_ids = [&job_id1, &job_id2];
+    /// let filter = MutationFilter::builder().jids(failed_ids.as_slice()).build();
+    /// client.requeue(MutationTarget::Retries, &filter).await.unwrap();
```
jonhoo, line 24: This, too, I have a hard time following — this is assuming that two jobs have _already_ failed (I think), and it tries to re-enqueue them _from_ the "retries" set? Why not from the dead set? And where do they get scheduled back — the queue they were originally on?

src/proto/client/mutation.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// # });
+    /// ```
+    pub async fn requeue<'a, F>(&mut self, target: MutationTarget, filter: F) -> Result<(), Error>
```
jonhoo, line 27: The argument name `target` makes me think that the jobs will be moved _to_ this, but I don't think that's the case. Isn't it more like `source` or `candidate_job_set` or something?

src/proto/single/mutation.rs
```diff
@@ -0,0 +1,112 @@
...
+#[builder(setter(into), build_fn(name = "try_build", private), pattern = "owned")]
+#[non_exhaustive]
+pub struct MutationFilter<'a> {
+    /// A job's [`kind`](crate::Job::kind).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "jobtype")]
+    #[builder(default)]
+    pub kind: Option<&'a str>,
+
+    /// Ids of jobs to target.
```
jonhoo, line 57:
```suggestion
    /// [`JobId`]s to target.
```

src/proto/single/mutation.rs
```diff
@@ -0,0 +1,112 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
```
jonhoo, line 58: Should it be send if empty?

src/proto/client/mutation.rs
```diff
@@ -0,0 +1,182 @@
...
+        &mut self,
+        target: MutationTarget,
+        jids: &'_ [&'_ JobId],
+    ) -> Result<(), Error> {
+        let filter = MutationFilter::builder().jids(jids).build();
+        self.mutate(MutationType::Discard, target, Some(&filter))
+            .await
+    }
+
+    /// Kil the jobs.
```
jonhoo, line 97:
```suggestion
    /// Kill a set of jobs.
```

src/proto/single/mutation.rs
```diff
@@ -0,0 +1,112 @@
...
+// will be taken into account, rather EITHER `jids` OR optional `kind`
+// plus optional `pattern`.
+// See: https://github.com/contribsys/faktory/issues/489
+//
+/// A filter to help narrow down the mutation target.
+///
+/// As of Faktory version 1.9.2, if [`MutationFilter::pattern`] and/or [`MutationFilter::kind`]
+/// is specified, the values in [`MutationFilter::jids`] will not be taken into account by the
+/// server. If you want to filter by `jids`, make sure to leave other fields of the filter empty
+/// or use dedicated methods like [`Client::requeue_by_ids`].
```
jonhoo, line 33: If that's the case, I wonder if we shouldn't derive builder and instead just have specific constructors (for now) for the combinations that _do_ work. What do you think?

jonhoo (changes requested):
Haven't gotten all the way through this, but racked up a few comments so figured I'd leave them with you!

src/lib.rs
rustworthy: added namespace and - indeed - prefix no longer needed in the public api

src/proto/single/cmd.rs
rustworthy: A [nil filter](https://github.com/contribsys/faktory/blob/5e1a0d938e4b19b8eff4d20e815289d536a1ae8c/server/mutate.go#L80) and an [empty one](https://github.com/contribsys/faktory/blob/5e1a0d938e4b19b8eff4d20e815289d536a1ae8c/server/mutate.go#L110) are treated in the common manner - as if we were matching _everything_  in the targeted queue (they are wildcards effectively). Added a comment.

I see what you mean, like what if the default behavior changes? well, as one put it, where possible, your interface should be intuitive enough that if the user has to guess, they usually guess correctly. Looking at the filter fields (jids, regexp, jobtypes) I am guessing that leaving those empty should be a considered a wildcard _or_ should cause an error encouraging me to be specific. In the same manner, omitting the filter - I am guessing - should be either a wildcard _or_ (yet again) should error back.

rustworthy: P.S.: to give an example of a broken law of least astonishment: in case of an unset filter, we  system would be targeting the jobs whose jobtype is a string literal "general" :cry:

src/proto/client/mutation.rs
mperham: The scheduled set is not a queue, it’s time-sorted data structure in Redis. The job is moved from that structure into its queue for a worker to fetch.

rustworthy: Actually they are not in queues, rather in dedicated sets where they are "pending" basically. 
Updated the docs and also added an example scenario

src/proto/single/mutation.rs
rustworthy: 1) the job sets are only used in mutation API context and also declared in the `mutate` module in the Go [bindings](5e1a0d938e4b19b8eff4d20e815289d536a1ae8c). As per another thread, we decided to remove the `Mutation` prefix from the constucts in the `mutation` namespace, so it's now simply `Target`. But I see what you mean this could be in theory simply `JobSet` or `JobSetKind`.

2) well, this where I was hesitant for a while, but decided to follow the Go [bindings way](https://github.com/contribsys/faktory/blob/5e1a0d938e4b19b8eff4d20e815289d536a1ae8c/client/mutate.go#L126). open to revisiting this place though

src/proto/client/mutation.rs
rustworthy: This is an outdated example, but yes it assumed that the jobs were failed but still had some retries attempts left, which is why they were pending in the `retries` set rather than `dead`. And yes they are being pushed back onto the queue they were initially destined for. Reminder: the queue name is encoded into the job itself with the "default" being a fallback value and so scheduling it on another queue would be _surprising_ breaking the rule of least astonishment. If we  had a property of the `Job` called `dead_letter_queue`, then yes there could be a fair doubt, like which queue we are now scheduling onto - still `Job::queue` of `Job::dead_letter_queue` already.

I am sorry you had a hard time following, but I were guessing exactly right :wink:

src/proto/client/mutation.rs
rustworthy: On the second thought, `candidate_job_set` looks more universal since we also have such methods as `Client::discard` and `Client::kill` and `Client::clear`

src/proto/single/mutation.rs
rustworthy: No need to. Even though is does not change the outcome, we should be consistent. So I added a utility trait called `Empty` (only for usage within the crate) so that we can do:
```rust
#[serde(skip_serializing_if = "Empty::is_empty")]
```

rustworthy: Also added a few unit tests for `Filter` and `MutationAction` for visualization purposes solely

src/proto/single/mutation.rs
rustworthy: I think, this is the right call. If I do not read the docs (which I definitely can) and specify both `kind` and `jids`  on the builder (thinking I am helping the server find jobs in question faster), I am getting into a situation where _all_ the jobs of that kind  in the targeted set are affected. If the server ignored  `pattern` and `kind` when `jids` are there (which is a trivial fix as Mike points out [here](https://github.com/contribsys/faktory/issues/489#issuecomment-2495112969), there would be no harm actually.

So I think the constructors will be:
```rust
Filter::by_ids;
Filter::by_kind;
Filter::by_pattern;
Filter::by_kind_and_pattern;
```
And the fields on that struct should be made private apparently.

rustworthy: went with `Filter::from_<attr>` rather than `Filter::by_<attr>`.  the latter is better for a verb I think, like `Client::filter_by_ids`.

src/lib.rs
rustworthy: though after some refactoring there are only two members left in that namespace :)

rustworthy:
> Haven't gotten all the way through this, but racked up a few comments so figured I'd leave them with you!

Sorry it took me a while to iterate on this PR - got some things to do in the `wewerewondering` project and also  was moving countries with my partner and cats (since we implement `Send`  :smile:  )

Please have a look at what we got.

src/proto/single/cmd.rs
jonhoo: Worth adding a comment here stating that that is the case then I think.

src/proto/client/mutation.rs
jonhoo: No no, all good! Thanks for your input here Mike :)
Updated docs + example helps :+1:

jonhoo:
I'm way behind on reviews again, sorry!

src/proto/single/cmd.rs
```diff
@@ -311,3 +313,71 @@ impl<'a, S: AsRef<str>> QueueControl<'a, S> {
...
     }
 }
+
+// ---------------------- MUTATE -------------------
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize)]
+#[serde(rename_all = "lowercase")]
+#[non_exhaustive]
+pub(crate) enum MutationType {
+    #[default]
```
jonhoo, line 323: Is it reasonable for this to even implement `Default`?

src/proto/single/cmd.rs
jonhoo: Ah, you already did!

src/proto/single/mutation.rs
```diff
@@ -0,0 +1,133 @@
...
+            kind: Some(job_kind),
+            ..Default::default()
+        }
+    }
+
+    /// Creates a filter with the provided search pattern.
+    ///
+    /// Faktory will pass this directly to Redis's `SCAN` command,
+    /// so please see the [`SCAN` documentation](https://redis.io/docs/latest/commands/scan/)
+    /// for further details.// Creates a filter with provided job kind.
```
jonhoo, line 98:
```suggestion
    /// for further details.
```

src/proto/single/utils.rs
```diff
@@ -152,3 +152,22 @@ mod test {
         assert_eq!(server, deserialized);
     }
 }
+
+pub(crate) trait Empty {
```
jonhoo, line 156: Ah, the trait here is so that this'll work for, say, `Option<Filter>`?

jonhoo (approved):
This is looking great now, thank you! Just a few very minor nits, and then we can get it in. Sorry again for the delay :)

src/proto/single/utils.rs
rustworthy: Yes, but also for this :
```rust
/// [`JobId`]s to target.
#[serde(skip_serializing_if = "Empty::is_empty")]
pub(crate) jids: Option<&'a [&'a JobId]>,
```

Not sealing it, because it's only for internal usage and is not leaking

rustworthy:
> This is looking great now, thank you! Just a few very minor nits, and then we can get it in. Sorry again for the delay :)

No worries, the event loop was not thrashed at all :wink:

src/proto/single/mutation.rs
```diff
@@ -0,0 +1,133 @@
...
+    Retries,
+
+    /// A set of failed jobs that will not be retried.
+    Dead,
+}
+
+// As of Faktory v1.9.2, not all the fields on the filter
+// will be taken into account, rather EITHER `jids` OR optional `kind`
+// plus optional `pattern`.
+// See: https://github.com/contribsys/faktory/issues/489
```
rustworthy, line 27: Jids should take upper hand, please see [PR](https://github.com/contribsys/faktory/pull/500)

jonhoo (approved):
Excellent, let's get this in, thank you!

Issue #88, opened 2024-11-19
Make `Failure` public


rustworthy:
Users should be able to read [`Failure`](https://github.com/jonhoo/faktory-rs/blob/6ff7d60f2cb370c0db59c423786571d0c21bb54f/src/proto/single/mod.rs#L208-L220)'s fields.

There is also a field in that structure not covered just yet called [`remaining`](https://github.com/contribsys/faktory/blob/main/client/job.go#L25), meaning the number of remaining retry attempts.

We will probably want to implement #61 first, so that we can "force-push" failed jobs from `retries` set to `sheduled` set, then 
 - in our test - consume the job again and examine it's `failure` field via `Job::failure` getter. Currently, we would have to wait for the Faktory server to do so or - alternatively - go to the Web UI and re-queue the job from there, which is - in both cases - not suitable for auto-tests.

rustworthy:
released in `v0.13.1`

PR #89, opened 2024-11-29
Job Failure

rustworthy:
Addresses #88

- - -
This change is [Reviewable](https://reviewable.io/reviews/jonhoo/faktory-rs/89)

### DIFF
--- a/src/proto/single/mod.rs
+++ b/src/proto/single/mod.rs
@@ -212,7 +212,7 @@ impl JobBuilder {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[non_exhaustive]
 pub struct Failure {
-    /// [`Number`](Job::retry) of times this job can be retried.
+    /// [Number](Job::retry) of times this job can be retried.
     pub retry_count: usize,
 
     /// Number of remaining retry attempts.
--- a/src/proto/single/mod.rs
+++ b/src/proto/single/mod.rs
@@ -297,9 +297,7 @@ impl Job {
     ///
     /// To access the entire failure, if any, use [`Job::failure`].
     pub fn failure_message(&self) -> Option<&str> {
-        self.failure()
-            .as_ref()
-            .and_then(|f| f.message.as_ref().map(|m| m.as_str()))
+        self.failure().as_ref().and_then(|f| f.message.as_deref())
     }
 }
 
--- a/src/proto/single/mod.rs
+++ b/src/proto/single/mod.rs
@@ -124,6 +124,8 @@ pub struct Job {
     /// Defaults to 25.
     #[serde(skip_serializing_if = "Option::is_none")]
     #[builder(default = "Some(JOB_DEFAULT_RETRY_COUNT)")]
+    // TODO: this should probably be a usize, see Failure::retry_count
+    // TODO: and Failure::retry_remaining. This can go to 0.14 release
     pub retry: Option<isize>,
 
     /// The priority of this job from 1-9 (9 is highest).
@@ -206,19 +208,39 @@ impl JobBuilder {
     }
 }
 
+/// Details on a job's failure.
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[non_exhaustive]
 pub struct Failure {
-    retry_count: usize,
-    failed_at: String,
+    /// [`Number`](Job::retry) of times this job can be retried.
+    pub retry_count: usize,
+
+    /// Number of remaining retry attempts.
+    #[serde(rename = "remaining")]
+    pub retry_remaining: usize,
+
+    /// Last time this job failed.
+    pub failed_at: DateTime<Utc>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    next_at: Option<String>,
+
+    /// When this job will be retried.
+    ///
+    /// This will be `None` if there are no retry
+    /// attempts (see [`Failure::retry_remaining`]) left.
+    pub next_at: Option<DateTime<Utc>>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    message: Option<String>,
+
+    /// Error message, if any.
+    pub message: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "errtype")]
-    kind: Option<String>,
+
+    /// Error kind, if known.
+    pub kind: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    backtrace: Option<Vec<String>>,
+
+    /// Stack trace from last failure, if any.
+    pub backtrace: Option<Vec<String>>,
 }
 
 impl Job {
--- a/src/proto/single/mod.rs
+++ b/src/proto/single/mod.rs
@@ -291,17 +291,6 @@ impl Job {
     pub fn failure(&self) -> Option<&Failure> {
         self.failure.as_ref()
     }
-
-    /// Error message for this job, if any.
-    ///
-    /// A convenience method that, internally, will check if this
-    /// job has got a [`Failure`] and - if it does - will access
-    /// the failure's message (see [`Failure::message`]).
-    ///
-    /// To access the entire failure, if any, use [`Job::failure`].
-    pub fn failure_message(&self) -> Option<&str> {
-        self.failure().as_ref().and_then(|f| f.message.as_deref())
-    }
 }
 
 pub async fn write_command<W: AsyncWrite + Unpin + Send, C: FaktoryCommand>(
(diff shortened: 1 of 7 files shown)


src/proto/single/mod.rs
```diff
@@ -206,19 +208,39 @@ impl JobBuilder {
     }
 }
 
+/// Details on a job's failure.
 #[derive(Serialize, Deserialize, Debug, Clone)]
+#[non_exhaustive]
 pub struct Failure {
-    retry_count: usize,
-    failed_at: String,
+    /// [`Number`](Job::retry) of times this job can be retried.
```
jonhoo, line 215:
```suggestion
    /// [Number](Job::retry) of times this job can be retried.
```

jonhoo, line 215: Are we sure that this is the number of times it _can_ be retried, as opposed to the number of times it _has_ been retried (which it feels like the name implies)

src/proto/single/mod.rs
```diff
@@ -206,19 +208,39 @@ impl JobBuilder {
...
+    pub retry_count: usize,
+
+    /// Number of remaining retry attempts.
+    #[serde(rename = "remaining")]
+    pub retry_remaining: usize,
+
+    /// Last time this job failed.
+    pub failed_at: DateTime<Utc>,
     #[serde(skip_serializing_if = "Option::is_none")]
```
jonhoo, line 224: This and the other `#[serde]` attributes seem misplaced now — they should come after the doc comments for the relevant field.

src/proto/single/mod.rs
```diff
@@ -206,19 +208,39 @@ impl JobBuilder {
...
-    message: Option<String>,
+
+    /// Error message, if any.
+    pub message: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "errtype")]
-    kind: Option<String>,
+
+    /// Error kind, if known.
+    pub kind: Option<String>,
```
jonhoo, line 239: Is kind actually an arbitrary string? Things like this is why `Failure` wasn't previously `pub`.

src/proto/single/mod.rs
```diff
@@ -266,6 +288,17 @@ impl Job {
     pub fn failure(&self) -> &Option<Failure> {
```
jonhoo, line 288: nit: this should probably return `Option<&Failure>`

src/proto/single/mod.rs
```diff
@@ -266,6 +288,17 @@ impl Job {
...
         &self.failure
     }
+
+    /// Error message for this job, if any.
+    ///
+    /// A convenience method that, internally, will check if this
+    /// job has got a [`Failure`] and - if it does - will access
+    /// the failure's message (see [`Failure::message`]).
```
jonhoo, line 296: I don't know if this convenience method is worthwhile?

tests/real/community.rs
```diff
@@ -795,46 +800,183 @@ async fn test_jobs_with_blocking_handlers() {
...
+    let local = "test_panic_and_errors_in_handler";
+    let mut c = Client::connect().await.unwrap();
+    let pattern = format!(r#"*\"args\":\[\"{}\"\]*"#, local);
+    c.requeue(
+        MutationTarget::Retries,
+        MutationFilter::builder().pattern(pattern.as_str()).build(),
+    )
+    .await
+    .unwrap();
+    c.queue_remove(&[local]).await.unwrap();
```
jonhoo, line 832: Why is this necessary?

tests/real/community.rs
```diff
@@ -795,46 +800,183 @@ async fn test_jobs_with_blocking_handlers() {
...
+    // ... consume all the jobs from the queue and _fail_ them
+    // "in different ways" (see our worker setup above)
+    for _ in 0..njobs {
+        assert!(w.run_one(0, &[local]).await.unwrap());
+    }
 
-    // we _did_ consume and process the job, the processing result itself though
-    // was a failure; however, a panic in the handler was "intercepted" and communicated
-    // to the Faktory server via the FAIL command;
-    // note how the test run is not interrupted with a panic
```
jonhoo, line 823: Feels worthwhile to preserve this comment in the new code.

tests/real/community.rs
```diff
@@ -795,46 +800,183 @@ async fn test_jobs_with_blocking_handlers() {
...
+    // let's now make sure all the jobs are re-enqueued
+    c.requeue(
+        MutationTarget::Retries,
+        MutationFilter::builder().pattern(pattern.as_str()).build(),
+    )
+    .await
+    .unwrap();
+
+    // now, let's create a worker who will only send jobs to the
+    // test's main thread to make some assertions
```
jonhoo, line 896: I'd like a description here about why we have to do this in two passes. I _think_ it's because the failure information in the job isn't something that's passed back to the client, it's something that's passed "forward" to the next handler the job is passed to? But that's not spelled out here.

tests/real/community.rs
```diff
@@ -881,11 +1026,43 @@ async fn mutation_requeue_jobs() {
         .await
         .unwrap();
 
-    // the job has been re-enqueued and we consumed it again
-    let had_one = worker.run_one(0, &[local]).await.unwrap();
-    assert!(had_one);
+    // ... this time, instead of failing the job this time, let's
+    // create a new woker that will just send the job
+    // to the test thread so that we can inspect and
+    // assert on the failure
```
jonhoo, line 1032:
```suggestion
    // assert on the failure from the first run
```

src/proto/single/mod.rs
rustworthy: hmm, but then why would we have `failure` accessor in our public API. 

I understand that breaking changes are unpleasant, but do you think the fact that the `kind` may become "enumerated"  in Faktory (even if this enumeration were not exhaustive)  tomorrow should prevent us from exposing `Failure` in the bindings today? Or you mean more research is needed on whether this can enumerated?

Let me double-check though

src/proto/single/mod.rs
rustworthy: Maybe we can nuke it, I will then use a helper function in the tests binaries

tests/real/community.rs
rustworthy: This clean up logic is needed because we are  (well at least I am) re-using the same container when developing locally. Without it, Faktory will re-schedule the failed jobs _after_ the test run, and so the next test run will face this remainders

src/proto/single/mod.rs
rustworthy: Sorry for this, must have tilted

src/proto/single/mod.rs
rustworthy: You are right and we are actually [demonstrating](https://github.com/jonhoo/faktory-rs/blob/feature/job-failure/tests/real/community.rs#L1054-L1058) this in the tests

tests/real/community.rs
rustworthy: threw the baby out with the bathwater

tests/real/community.rs
rustworthy: Exactly, and it is also available via Web UI. Add t a comment:
```
we _did_ consume and process the job, the processing result itself though
was a failure; however, a panic in the handler was "intercepted" and communicated
to the Faktory server via the FAIL command, the error message and the backtrace, if any,
will then be available to a Web UI user or to the worker consumes the retried job
(if `Job::retry` is Some and > 0 and there are `Failure::retry_remaining` attempts left)
in this job's `failure` field; see how we are consuming the retried jobs later
in this test and examin the failure details;
```

src/proto/single/mod.rs
rustworthy: In our codebase, the kind (`errtype` in Faktory) is always an "unknown" literal, we construct a `Fail` of this kind with `Fail::generic` all the time. 

Also "unknown" is what they use in [go bindings](https://github.com/contribsys/faktory/blob/10ccc2270dc2a1c95c3583f7c291a51b0292bb62/client/client.go#L363). Notice how this is actually hard-coded.

I wish we could enumerate this (with a non-exhaustive enum where `Unknown` would be the only variant for now), but what if there are also other workers (in other languages or not using the `faktory-rs` bindings), how do we unmarshal that? 

___

The more I am thinking about it, the more I am questioning if there is any usefulness in this field at all.
We can not make `kind` public for now at all, and if in the future it is no longer a hard-coded "unknown" literal, we could consider enumerating it, wdyt?

jonhoo: Yeah, I like that idea — let's just make that field not `pub` for now!

tests/real/community.rs
jonhoo: Would you mind leaving a slightly more elaborate comment to that effect?

rustworthy: Sure thing

jonhoo (approved):
Approving subject to the last two open threads which have easy fixes :+1:

